PR #148, opened 2024-05-13
refactor(backend/mixins): make AggregateQuerysetMixin more readable

c0rydoras:
prefetch_related_for_field was dropped because it isn't used anywhere


winged (approved):
Nice work cleaning this up!

PR #170, opened 2024-05-15
fix(backend): move conftest.py out of wheel

c0rydoras:


hairmare (approved):
such fixture, much nice

PR #180, opened 2024-05-16
fix: error notification when running onboarding tours

### LINKED ISSUE #144: [BUG]: Tour yields two "unexpected error" notifications
### Description

When doing the tour, two "unexpected error" notifications happen

### Steps to Reproduce

1. start dev setup
```bash
git clone https://github.com/adfinis/timed
cd timed
make start
make keycloak-import
make loaddata
```
2. head to https://timed.local and log in (user doesn't matter)
3. Press continue/next buttons on tour
4. receive "unexpected error"

### Expected Behavior

No "unexpected error"

### Attachments

(in chronological order)

image
image
image
image
image

### Additional Context

_No response_

### Version

v4.0.1

### Required

MitanOmar:
fix #144

### DIFF
--- a/frontend/app/services/tour.js
+++ b/frontend/app/services/tour.js
@@ -159,8 +159,8 @@ export default class TourService extends Tour {
               await this.router.transitionTo(
                 this.autostartTour.undoneTours.shift() ?? "index"
               );
-            } catch (error) {
-              this.notify.error("unexpected error");
+            } catch {
+              /* eslint:disable:no-empty */
             }
           }
         };


frontend/app/services/tour.js
```diff
@@ -159,8 +159,8 @@ export default class TourService extends Tour {
               await this.router.transitionTo(
                 this.autostartTour.undoneTours.shift() ?? "index"
               );
-            } catch (error) {
-              this.notify.error("unexpected error");
+            } catch (_) {
+              _.toString();
             }
```
c0rydoras, line 164: this:

1. doesn't fix the underlying issue
2. the `_.toString()` doesn't do anything and is not necessary

Since this is important for onboarding tho, we can probably use this as a **temporary** solution
```suggestion
            } catch {}
```

(i haven't tested the suggestion but it should be equivalent to the proposed **temporary** solution

c0rydoras (changes requested):

frontend/app/services/tour.js
c0rydoras: i haven't done frontend in a while, you can probably just add a **temporary** eslint ignore comment

c0rydoras: But yes, eslint is right, we shouldn't have empty block statements

c0rydoras: AFAIK, especially in try catch statements, an empty (catch) block is an antipattern

c0rydoras: @derrabauke what would you say?

derrabauke: Why fix the symptom and not the root cause? @MitanOmar, why is the transition aborted in the first place?

MitanOmar: @derrabauke i really don't know, but when i put `this.router.transitionTo` inside any try catch, even not there, it's throwing this error

MitanOmar: @derrabauke for example, in the home page, i change the routing from
```hbs
          <LinkTo @route="index.attendances">
            Attendances
            <span class="badge">{{humanize-duration this.attendanceSum}}</span>
          </LinkTo>
```
to be an action in the controller with `this.router.transitionTo` with try catch, it was throwing the same error again

derrabauke: I've checked it. Just remove the `try-catch` block. It's not necessary at this point.

MitanOmar: @derrabauke done

derrabauke: Sure you pushed the changes? O.o

MitanOmar: yes, this is the final results :)

derrabauke: image

the catch block has still some non-sense code in it. Would expect `catch {}` with some eslint exception comment in it!?

MitanOmar: @derrabauke  it's empty now

derrabauke (approved):

frontend/app/services/tour.js
```diff
@@ -159,8 +159,8 @@ export default class TourService extends Tour {
               await this.router.transitionTo(
                 this.autostartTour.undoneTours.shift() ?? "index"
               );
-            } catch (error) {
-              this.notify.error("unexpected error");
+            } catch {
+              /* tslint:disable:no-empty */
```
c0rydoras, line 163: tslint? AFAIK we use eslint, and no typescript :c

MitanOmar: changed, Thanks

derrabauke: true dat.. `// eslint-disable-next-line no-empty`

c0rydoras (approved):

frontend/app/services/tour.js
```diff
@@ -159,8 +159,8 @@ export default class TourService extends Tour {
               await this.router.transitionTo(
                 this.autostartTour.undoneTours.shift() ?? "index"
               );
-            } catch (error) {
-              this.notify.error("unexpected error");
+            } catch {
+              /* eslint:disable:no-empty */
```
derrabauke, line 163:
```suggestion
              // eslint-disable-line no-empty
```

Issue #90, opened 2023-05-22
Enhancement: extra sentence in "30% and 70% budget check" message


rafahu:
Hi

Currently 2 messages are being generated automatically by Timed once 30% and 70% of the total budget has been reached. The text is
```
## Project exceeded 30% of budget

- Billable Hours: $total hours in time
- Budget: $total budget

To PM: Please check the remaining effort estimate. If more budget is needed, reach out to relevant stakeholders.
```
and 
```
## Project exceeded 70% of budget

- Billable Hours: $total hours in time
- Budget: $total budget

To PM: Please check the remaining effort estimate. If more budget is needed, reach out to relevant stakeholders.
```

I would like to change the bottom section of the message **in both** to

```
To PM: 
- Please check the remaining effort estimate. If more budget is needed, reach out to relevant stakeholders.
- Make a Redmine entry with your budget review conclusions.
```

c0rydoras:
@rafahu @trowik whats the current state of this?

trowik:
Text seems to be updated. Can we close it @rafahu ?

rafahu:
Yes @trowik I saw the comment in 2 different tickets. Good to close. Thanks.

PR #190, opened 2024-05-21
chore: update budget reminder text

trowik:
Ref #90


c0rydoras:
AFAIK This shouldn't be a `chore`, it changes how the app works and a chore won't create a new release, maybe use a `fix` or a `feat` as scope

c0rydoras (approved):
looks good

Issue #142, opened 2024-05-13
Move record to another project fails to fetch full project list


shurik-io:
**Describe the bug**
During verification I need to move records another customer or project, as they were reported on the wrong position. So I select the records I need and click "Edit" and then try to choose to correct project, but cannot select it, as it didn't fetch all available project within the same customer.

Other users reported, to switch customer to a random other one, switch back the required customer and then the list is loaded correctly. Sadly I couldn't reproduce the error 5d later (long weekend). But at time of happening, it was reproducible.

**To Reproduce**

1. Select records to move
2. Click Edit
3. Select project drop-down, 
4. Not the full list is shown with all available projects of that customer.

Refreshing the browser page (CTRL+F5) before clicking "Edit" didn't help either.

**Expected behavior**
Full project list is shown

**Screenshots**
![image](https://github.com/adfinis/timed-frontend/assets/35893671/9204d6dc-2989-4099-8ccd-9ed7d23ed7e8)

**Desktop (please complete the following information):**

- OS: `Linux hostname 6.8.8-200.fc39.x86_64 adfinis/timed-frontend#1 SMP PREEMPT_DYNAMIC Sat Apr 27 17:42:13 UTC 2024 x86_64 GNU/Linux`
- Browser: Mozilla Firefox
- Version: 125.0.3

PR #207, opened 2024-05-23
fix(frontend): `TaskSelection` component is not loading the full customer project in the first page load

MitanOmar:
fix: #142


c0rydoras:
This can probably be done by adjusting the `include` when initially fetching the project

c0rydoras:
Which is better as then we don't do more requests

frontend/app/components/task-selection/component.js
```diff
@@ -116,6 +136,13 @@ export default class TaskSelectionComponent extends Component {
     } else {
       this.tracking.fetchCustomers.perform();
     }
+
+    if (this.customer?.id) {
+      this.customer.project = this.customerProjectsData.value ?? [];
+    }
+    if (this.project?.id) {
+      this.project.tasks = this.projectTasksData.value ?? [];
+    }
```
derrabauke, line 145: @MitanOmar Is this copy :spaghetti:  from the `tracking` service? 
This duplicates requests and does not respect what is done in the code above. I think an include as proposed by Arthur or a fix in the `onCustomerChange` method is necessary. Why is this error surfacing and under what circumstances (please don't quote the issue)?

c0rydoras:
ill close this as #142 has been fixed by #447

PR #210, opened 2024-05-23
chore(github-workflow): run the tests only when the tests belong to t…

MitanOmar:
…he changed folder


.github/workflows/chart.yaml
```diff
@@ -4,8 +4,12 @@ name: Chart
 on:
   push:
     branches: [main]
+    paths:
+      - 'charts/**'
```
hairmare, line 8: the chart tests also build and install the container images so these tests could in theory also fail if we change things in other paths

as such they currently seem to serve as simple full stack smoketests, maybe we want to consider keeping them on all changes for that purpose?

MitanOmar: i remove the condition from the charts workflow, but why i see that the tests are still `Waiting for status to be reported`

MitanOmar: @hairmare do you know why there is three frontend tests are there even if we define the job to be executed only if there is a change in the `frontend` folder ?

c0rydoras: our rules still need them to succeed

c0rydoras: image

MitanOmar: there is no way to define the required jobs to be defined in the file ?

c0rydoras: there are probably ugly workarounds

MitanOmar: i will keep this PR open, since there are other tasks more important than this, we can come to it later

c0rydoras:
```yaml
# without paths
...
jobs:
  dirty:
     outputs:
       dirty: ${{ steps.check-dirty.outputs.dirty }}
     steps:
      - id: dirty
        run: |
          # bash script to determine if dirty, then write to output

  lint:
      needs: dirty
      steps:
       - name: foo
         if: ${{ needs.dirty.outputs.dirty }}
         run: |
           ls -a
```

something like this would probably work, but is very ugly

MitanOmar: i totally agree that it's ugly

derrabauke: I'm not sure if it works, but could you try the inverse of `paths`.
* For the **backend** workflow:
```yml
  pull_request:
    paths-ignore:
      - 'frontend/**'
```
* For the **frontend** workflow:
```yml
  pull_request:
    paths-ignore:
      - 'backend/**'
```

derrabauke (approved):

.github/workflows/backend.yaml
```diff
@@ -4,7 +4,15 @@ name: Backend CI
 on:
   push:
     branches: [main]
+    paths-ignore:
```
derrabauke, line 7: As already discussed, let's use the positive `path` matching.

derrabauke (changes requested):
As already discussed, let's use the positive path matching. CI rules got adapted in the meantime.

.github/workflows/backend.yaml
MitanOmar: just remove my last commit, and it's now erady

PR #221, opened 2024-05-27
fix(frontend): hide task form after creating new project task

### LINKED ISSUE #44: Projects UX: Creating multiple new Tasks
## Momentanes Verhalten

Currently, the newly created task is opened in a state where it can be edited. In order to create another task, "New Task" needs to be selected again. Users not paying close attention are more likely to change the newly created task instead of adding another one.
## Erwartetes Verhalten

Adding multiple tasks should be less confusing, either by

-    hiding the form after adding a task, or
-    show the real "add" form after adding a task

I think I'd prefer option one.
## Schritte zum Reproduzieren

-    Hit "New Task" in "projects"
-    Save the new task
-    Task is opened in "edit" state

MitanOmar:
fix: #44


frontend/tests/acceptance/project-test.js
```diff
@@ -80,9 +82,6 @@ module("Acceptance | projects", function (hooks) {
 
     await click("[data-test-save]");
 
-    assert.dom("[data-test-name]").hasValue("FooBar Task 1");
-    assert.dom("[data-test-reference]").hasValue("Reference of FooBar Task 1");
-    assert.dom("[data-test-estimated-time]").hasValue("02:15");
```
derrabauke, line 85: Let's test for this stuff later in the flow. Simply save the inputs and then click on the created task again.

Issue #228, opened 2024-05-30
[BUG]: Projects can't be created in django admin


c0rydoras:
### Description

Can't create projects in django admin

### Steps to Reproduce

1. open django admin
2. click projects
3. receive error

### Expected Behavior

no error

### Attachments

![image](https://github.com/adfinis/timed/assets/110528300/6dbd8a2e-2ebd-42ee-b5fc-799cbe54e9c2)

stacktrace
```
Environment:


Request Method: GET
Request URL: https://timed.local/admin/projects/project/add/

Django Version: 4.2.11
Python Version: 3.12.2
Installed Applications:
['timed.apps.TimedAdminConfig',
 'django.contrib.humanize',
 'multiselectfield.apps.MultiSelectFieldConfig',
 'django.forms',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'hurricane',
 'rest_framework',
 'django_filters',
 'djmoney.apps.MoneyConfig',
 'mozilla_django_oidc',
 'django_prometheus.apps.DjangoPrometheusConfig',
 'corsheaders',
 'nested_inline',
 'timed.employment',
 'timed.projects',
 'timed.tracking',
 'timed.reports',
 'timed.redmine',
 'timed.subscription',
 'timed.notifications',
 'django_extensions']
Installed Middleware:
['django_prometheus.middleware.PrometheusBeforeMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django_prometheus.middleware.PrometheusAfterMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware']



Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
[...]

PR #229, opened 2024-05-30
fix(backend): projects can't be created in django-admin

c0rydoras:
fixes: https://github.com/adfinis/timed/issues/228


trowik (approved):
:+1:

PR #243, opened 2024-06-05
fix(frontend): update remaining effort when changing estimated time

### LINKED ISSUE #179: [BUG]: creating a new task when remaining effort is active
### Description

While creating a new task on a project with remaining effort tracking active and updating the "Estimated time" the "Remaining effort" value updates the field according to the "Estimated time" value.

When updating the "Estimated time" field by entering a value, "Remaining effort" takes that value.
When using arrows to update the "Estimated time" it only updates "Remaining effort" after the first click.
The update of "Remaining effort" only happens on the first update of the field "Estimated time".

### Steps to Reproduce

1. Go to "Projects" page
2. Activate "Remaining effort tracking" on a project
3. Click "Add Task"
4. Update value of "Estimated time" either by entering a value or using arrow keys

### Expected Behavior

"Remaining effort" value shouldn't update automatically or should always update automatically, not only on the first update of "Estimated time".

### Attachments

_No response_

### Additional Context

_No response_

### Version

v3.5.0 (frontend)

### Required

MitanOmar:
fix: #179


frontend/app/projects/controller.js
```diff
@@ -147,7 +147,7 @@ export default class ProjectsController extends Controller {
   updateRemainingEffort(changeset) {
     if (
       changeset.get("project.remainingEffortTracking") &&
-      !changeset.get("mostRecentRemainingEffort")
+      this.selectedTask.isNew
```
derrabauke, line 150: What is the value of `changeset.get("mostRecentRemainingEffort")` in case `this.selectedTask.isNew === true`?

PR #280, opened 2024-06-18
Fix/backend/drop multiselect field dependency

c0rydoras:


winged (approved):
LGTM - weird that the module still seems maintained on github, but no release in 4 years..

PR #282, opened 2024-06-18
fix(backend): correctly show sums in workreport

c0rydoras:
changes: removed 0s from backend/timed/reports/templates/workreport.ots


winged (dismissed):
Code LGTM, but can we have a test for that?

c0rydoras:
I could probably drop the code changes. The only relevant changes are the ones in the template:

image
image
image

Required Change:
image

open-dynaMIX (approved):
Great work :+1:

Issue #29, opened 2024-03-11
full text search for comments in analysis


derrabauke:
**Is your feature request related to a problem? Please describe.**
When filtering for a certain task, I want to be able to filter the results by certain comments made for the task. This is relevant for task which host a variable amount of smaller subtasks e.g. *Software Maintenance*

**Describe the solution you'd like**
In the *Analysis* and *Statistics* filter I want to be able to search for comments containing specific words. Ideally this would support multiple strings separated by commans and/or regular expressions.

**Additional context**
Maybe add the filter input here after the `task` select.
![image](https://github.com/adfinis/timed-frontend/assets/10029904/39d8e910-6b4e-4ab5-8fc8-9b9ddd86f2fa)

hairmare:
does it have to be regex or is it also time to look into natsearch algos again?

my feeling is that our current psql stack might enable some things that could make some nat-searching possible without much effort

this could be worth more than enabling regex (which i as a techie would still also like)

c0rydoras:
Another question is, do we implement it in the backend or the frontend?

## backend filter
- would be better, as there should be only as much logic in the frontend as required
- not sure how to implement
- new ReDoS vector

## frontend
- slow
- ...

derrabauke:
*Oh fo sho* in the backend. Not sure if we already use the full text search feature. Further there is [regex](https://docs.djangoproject.com/en/dev/ref/models/querysets/#regex) for querysets, but not sure about the value/ReDos ratio. :man_shrugging:

PR #283, opened 2024-06-18
feat: comment search

c0rydoras:
implements: https://github.com/adfinis/timed/issues/29


winged:
Quick question: how will this be used? As you note yourself, this may cause performance problems, not only for malicious queries (which I think we can neglect here) but also for regular users who don't know when a RE query will cause the DB to break down

winged:
> @winged #29 describes that

Thanks! Still a small explanation in the PR (and commit message!) won't hurt, along with a link in the original comment to the issue. Just to make things easier when we need to dig for reasons in 5 years

c0rydoras:
@winged i think we could use the `search` [lookup](https://docs.djangoproject.com/en/5.0/ref/contrib/postgres/search/)

c0rydoras:
I switched to postgres full text search instead of regex, this should make it so that ReDoS is no longer an issue, and it's probably more user friendly for nontechnical users :)

open-dynaMIX (changes requested):
Backend change looks good :+1:

However, it would be cleaner if this PR just contained two commits:

 - feat: added full text search for comments
 - ci(frontend): use pnpm v9

frontend/app/components/task-selection/component.js
```diff
@@ -49,6 +49,15 @@ export default class TaskSelectionComponent extends Component {
   @localCopy("args.initial.task")
   _task;
 
+  /**
+   * The manually written comment regex
```
open-dynaMIX, line 53: Still mentioning `regex`.

frontend/app/components/task-selection/component.js
```diff
@@ -214,6 +224,16 @@ export default class TaskSelectionComponent extends Component {
       : this._task;
   }
 
+  /**
+   * The comment regex
```
open-dynaMIX, line 228: Same here.

open-dynaMIX (changes requested):
Well done, with the commit cleanup :+1:

I found two references to `regex` in the changes.
After that, we just need a frontend approval :tada:

winged:
I much prefer the FTS search! However this will need an appropriate index on the `comment` field, otherwise it's gonna be rather slow, as it needs a full table scan each time, same as the regex lookup would have.

Here's how you can do that: https://docs.djangoproject.com/en/5.0/ref/contrib/postgres/search/#performance

frontend/app/components/task-selection/template.hbs
```diff
@@ -48,6 +48,13 @@
         optionTemplate=this.taskOptionTemplate
       )
     )
+    comment=(component
+      (ensure-safe-component "comment-input")
+      placeholder="Comment"
+      value=this.comment
+      onChange=this.onCommentChange
+      class="form-control comment-field"
+    )
```
derrabauke, line 57: I think the `comment-input` should be part of the `filter-sidebar` component, instead of the `task-selection`. It's more of a *filter* and the `task-selection` is used in various places. Also makes the whole initialization stuff easier, as you don't have to struggle with the `task-selection` component (which is *horrifying IMHO*)... :)

c0rydoras: i moved it to `filter-sidebar/filter` https://github.com/adfinis/timed/pull/283/commits/89f2bf1eaf68795e16bc54fe3b22ead5d1e14d15#diff-cdc879d2c92851bda53f78bd96507f73ea47da51b6f9c084e65bc0ca4fe09d48

derrabauke: Sorry for being so picky here, but I think this splintered component (previously `comment-input`) is not well maintainable. :sweat_smile: 

I wanted to suggest decoupling the whole comment search input from the `task-selection` component. Just like you started out in the beginning, with your `comment-input`, but instead of loading it into the task-selection component why not directly use it in the `analytics/index` template like we do for the [user select](https://github.com/adfinis/timed/commit/89f2bf1eaf68795e16bc54fe3b22ead5d1e14d15#diff-12c469691b8d9581a646f849b70514cb098e3442b584edecf96d2b6f9ac75079R60)? 

The `task-selection` component should really not become any more bloated as it already is.

So:
* (re-)add `comment-input` or whatever you wanna call it
* make direct use of it in `analysis/index` template
* keeps `filters-sidebar/filter` and `task-selection` untouched --> less :spaghetti: more :sparkles:

c0rydoras: yea i agree, i changed it now

derrabauke (approved):

open-dynaMIX (approved):

PR #311, opened 2024-06-26
fix(backend/activities): end activities when they get transferred

### LINKED ISSUE #233: [BUG]: You do not have permission to perform this action when stopping an activity
### Description

I'm unable to stop an activity that i started earlier today.

### Steps to Reproduce

1. be me
2. start activity
3. be done with working
4. click stop button next to running activity
5. receive error

### Expected Behavior

It stops the activity

### Attachments

I see the following error in the response to a PATCH req to https://timed.adfinis.com/api/v1/activities/214087
```json
{
  "errors": [
    {
      "detail": "You do not have permission to perform this action.",
      "status": "403",
      "source": {
        "pointer": "/data"
      },
      "code": "permission_denied"
    }
  ]
}
```

### Additional Context

I generated HAR if need be.

It's not logging much on the backend:

```
Forbidden: /api/v1/activities/214087
403 PATCH /api/v1/activities/214087 (10.100.18.71) 204.12ms
```

### Version

4.1.9

### Required

c0rydoras:
fix #233

### DIFF
--- a/backend/timed/tracking/serializers.py
+++ b/backend/timed/tracking/serializers.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING
 from django.contrib.auth import get_user_model
 from django.db.models import BooleanField, Case, Q, When
 from django.utils.duration import duration_string
+from django.utils.timezone import datetime
 from django.utils.translation import gettext_lazy as _
 from rest_framework_json_api import relations, serializers
 from rest_framework_json_api.relations import ResourceRelatedField
@@ -67,6 +68,20 @@ class ActivitySerializer(ModelSerializer):
             raise ValidationError(_("An activity block may not end before it starts."))
 
         return data
+
+    def update(
+        self, instance: models.Activity, validated_data: dict
+    ) -> models.Activity:
+        """Update an activity.
+
+        Ensure that transferring an activity ends it.
+        """
+        to_time = validated_data.get("to_time", instance.to_time)
+
+        if validated_data.get("transferred", instance.transferred) and not to_time:
+            validated_data["to_time"] = datetime.now().time()
+
+        return super().update(instance, validated_data)
 
     class Meta:
         """Meta information for the activity serializer."""
--- a/backend/timed/tracking/tests/test_activity.py
+++ b/backend/timed/tracking/tests/test_activity.py
@@ -355,3 +355,33 @@ def test_activity_set_to_time_none(internal_employee_client, activity_factory):
 
     res = internal_employee_client.patch(url, data)
     assert res.status_code == status.HTTP_400_BAD_REQUEST
+
+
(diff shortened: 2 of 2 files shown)


backend/timed/tracking/serializers.py
```diff
@@ -68,6 +69,21 @@ def validate_running_activity():
...
         return data
 
+    def update(
+        self, instance: models.Activity, validated_data: dict
+    ) -> models.Activity:
+        """Update an activity.
+
+        Ensure that transferring an activity ends it.
+        """
+        transferred = validated_data.pop("transferred", instance.transferred)
```
winged, line 79: Honest question: What even is a transferred activity? And why would we `pop()` the value here? This would imply that we'll never set that attribute again (as it's not in `validated_data` anymore after this line)

c0rydoras: AFAICT the frontend transfers an activity but leaves it running resulting in https://github.com/adfinis/timed/issues/233

c0rydoras: an activity gets set to transferred when you e.g. generate a timesheet, after that you can no longer edit it

c0rydoras:
```py
    permission_classes = (
        (
            # users may not change transferred activities
            IsAuthenticated & IsInternal & IsNotTransferred
            | IsAuthenticated & IsReadOnly
            # only external employees with resource role may create not transferred activities
            | IsAuthenticated & IsExternal & IsResource & IsNotTransferred
        ),
    )
```

winged: > an activity gets set to transferred when you e.g. generate a timesheet, after that you can no longer edit it

That makes sense, thanks! Found the permission code, but not what is causing it.

In that case, you probably still shouldn't `pop()` the value tho (use `.get()` instead)

backend/timed/tracking/tests/test_activity.py
```diff
@@ -355,3 +355,32 @@ def test_activity_set_to_time_none(internal_employee_client, activity_factory):
...
+    url = reverse("activity-detail", args=[activity.id])
+    response = internal_employee_client.patch(url, data)
+    assert response.status_code == status.HTTP_200_OK
+
+    json = response.json()
+
+    assert json["data"]["attributes"]["to-time"]
+
+    activity.refresh_from_db()
+    assert activity.to_time
```
winged, line 386:
```suggestion
    assert activity.to_time
    assert activity.transferred
```

winged (changes requested):

backend/timed/tracking/tests/test_activity.py
c0rydoras: done

winged (approved):

PR #315, opened 2024-06-28
ember 5.4

c0rydoras:


frontend/package.json
```diff
@@ -46,7 +48,6 @@
     "@glimmer/component": "1.1.2",
     "@glimmer/tracking": "1.1.2",
     "@html-next/vertical-collection": "4.0.2",
-    "@sentry/ember": "^7.98.0",
```
derrabauke, line 49: Why remove sentry? It was the goal to enable it some time.

c0rydoras: i did some local testing, i split these changes into seperate PRs

c0rydoras:
`ember-data` v5.3 is going to be painful

PR #320, opened 2024-06-28
fix(frontend): drop resize observer polyfill

c0rydoras:


c0rydoras:
AFAICT we can drop this unless we really need to support IE

derrabauke (approved):

derrabauke (approved):

derrabauke (changes requested):
Magic-Link seems broken by the change of `<Input />` to `<input />` :man_shrugging:

backend/ruff.toml
```diff
@@ -100,13 +100,12 @@ ignore = [
 ]
 
 [lint.isort]
-known-first-party = ["timed"]
+known-first-party = ["backupcontrol"]
```
winged, line 103: HWAT

backend/ruff.toml
```diff
@@ -70,7 +70,7 @@ ignore = [
 	"E501",    # managed by formatter
 	"TD002",   # don't require author of TODO
 	"TD003",   # don't require link to TODO
-	"D100",    # don't enforce existance of docstrings
+	"D100",    # don't enforce just do the same as here existance of docstrings
```
winged, line 73: HWAT

winged (changes requested):

backend/ruff.toml
c0rydoras: ah

backend/ruff.toml
c0rydoras: this doesn't belong here

derrabauke (approved):

winged (approved):

PR #331, opened 2024-07-08
fix(frontend): extract parse-filename into function and remove surrou…

c0rydoras:


MitanOmar (approved):

derrabauke:
:sparkles: nice :sparkles:

PR #353, opened 2024-07-17
chore(frontend): new deprecation to prepare fixing it before next emb…

MitanOmar:
…er upgrade


derrabauke:
- What does this "fix"? --> The commit message & PR title are not informative
- Why adding all those silencers?

MitanOmar:
at the moment, the reviewer can not verify reports because some of those deprecation are causing error.
so i include them here, and i will work on it one by one

PR #403, opened 2024-08-13
fix(frontend): adjust naming

c0rydoras:
fix: #54


trowik (approved):
:+1:

MitanOmar (approved):

trowik:
Don't know who created that issue, as it's from Redmine, but let's just go with it.
I don't necessarily see the need to adjust it, but it's fine

Issue #152, opened 2024-05-14
[BUG]: magic link button is only rounded on one side when in magic link modal


c0rydoras:
### Description

When in the magic link ui/when the magic link modal is opened the button is no longer rounded on all sides

### Steps to Reproduce

1. go to timed instance
2. click magic link button
3. its no longer rounded on all sides

### Expected Behavior

It to stay round 

### Attachments

![image](https://github.com/adfinis/timed/assets/110528300/f457b75f-0879-41d7-ba5f-c580712882b9)
![image](https://github.com/adfinis/timed/assets/110528300/3cad7f27-1f82-4bd3-9456-af24fbdac881)

### Version

v4.0.3

### Required

- [X] I have checked past issues to see if this bug has already been reported.

c0rydoras:
How is this not a bug?

c0rydoras:
What @derrabauke said should fix it

PR #405, opened 2024-08-13
fix(frontend/magic-link): button is now round again when modal is opened

c0rydoras:
fix: #152


c0rydoras:
ugly solution but it works.

MitanOmar (approved):

PR #412, opened 2024-08-14
chore(frontend): drop commitlint

c0rydoras:


derrabauke:
Why?

c0rydoras:
because it isn't used anymore ^^

derrabauke:
Obviously :man_facepalming: sorry. But is there an alterantive in place? `pre-commit`?

c0rydoras:
not really, pre-commit was also dropped at some point iirc

derrabauke:
> not really, pre-commit was also dropped at some point iirc

I would strongly suggest to set this up again. Saves a lot of headache when using sem-ver releases with auto changelog generation.

c0rydoras:
yea but we don't need it in our frontends dependencies

c0rydoras:
> I would strongly suggest to set this up again. Saves a lot of headache when using sem-ver releases with auto changelog generation.

why? our changelog doesn't look at that bad? https://github.com/adfinis/timed/releases

derrabauke:
> > I would strongly suggest to set this up again. Saves a lot of headache when using sem-ver releases with auto changelog generation.
> 
> why? our changelog doesn't look at that bad? https://github.com/adfinis/timed/releases

I does cause you adhere to the commitlint conventions. But imagine a junior dev joining the team who doesn't know. ;)

derrabauke (approved):
Anyways, it's the right strategy to remove the dep here.

PR #414, opened 2024-08-14
chore(frontend/deps): use ember-data v4.12

c0rydoras:


frontend/app/index/activities/controller.js
```diff
@@ -145,9 +145,7 @@ export default class ActivitiesIndexController extends Controller {
    */
   @action
   generateReportsCheck() {
-    const hasUnknown = !!this.activities.find(
-      (a) => a.task.get("id") === undefined
-    );
+    const hasUnknown = !!this.activities.find((a) => !a.task.get("id"));
```
c0rydoras, line 148: this no longer worked because the tasks id's were `null` and not `undefined`

frontend/app/index/reports/controller.js
```diff
@@ -96,6 +97,7 @@ export default class IndexReportController extends Controller {
 
   @action
   async createEmptyReport() {
+    if (!this.checkForEmptyRecord) return;
```
c0rydoras, line 100: without this we run into a race condition and the magic-link record and an empty record will be created

image

frontend/app/analysis/index/template.hbs
```diff
@@ -22,9 +23,6 @@
         @onReset={{queue t.clear this.reset}}
         as |fs|
       >
-        {{#if this.prefetchData.isRunning}}
```
derrabauke, line 25: Wanna re-add this in a unless/else block or is this not of value?

frontend/app/index/reports/route.js
```diff
@@ -58,5 +58,6 @@ export default class IndexReportsRoute extends Route {
         this.notify.error("Could not create report");
       }
     }
+    controller.checkForEmptyRecord = true;
```
derrabauke, line 61: What's that about?

frontend/app/analysis/index/template.hbs
c0rydoras: it is of value, without this `@initial` gets set before the the data is prefetched, which then causes this:

image

frontend/app/index/reports/controller.js
derrabauke: Thanks for the hint! Now I get it ;-)

frontend/app/analysis/index/template.hbs
c0rydoras: the problem is that `@initial` gets set before the data is fetched, which results in this:

image

Issue #79, opened 2023-09-05
Magic link workflow improvements


winged:
The magic links (#1022) are indeed magic! However, I wish for some changes that would massively improve the feature's usability and effectiveness.

Similar to adfinis/timed-frontend#685, I'd like to have the possibility to have a magic link anywhere (At minimum in activities and timesheet, possibly other places).

I kinda imagine it to work as follows:

* Clicking "new magic link" opens the dialog in an empty state (as it is now)
* Clicking "new magic link" on a timesheet/activity entry opens the dialog with already-populated values from corresponding entry
* The dialog is used both when receiving a link and when sending a link: Receiving a link opens the dialog just as if it was opened via button
* The dialog gets two additional buttons: "Start Tracking", and "Create timesheet entry".
* Add a date field as well, so we can create entries for past (or even future) days

Here's how I imagine it to look (roughly):

![image](https://github.com/adfinis/timed-frontend/assets/303759/fa12bc68-3d72-40d5-a357-63947f6c8f3f)

This would enable the following use cases:

* Invite someone to work on a task that I've already worked on (timesheet) or am tracking via activity
* When receiving a magic link decide whether to start tracking or create a timesheet entry directly
* The "resume" button on the activity could be entirely replaced with the magic link dialog as well

c0rydoras:
@winged whats the current state of this?

also #136 is related right?

winged:
Yes, I even think #136 can be closed in favour of this one, as this is more detailled.

There has not been any (serious) work done on this, the screenshot above is just a mockup of mine to illustrate the behaviour

PR #417, opened 2024-08-14
magic link improvements

c0rydoras:
Improve magic links by making it so just one of comment/duration/customer/project is enough to create a report

This is a draft for the following reasons:
- :spaghetti: code
- buggy (after saving a report created by a magic link that didn't have a task, it gets set to blank in the frontend)

### DIFF
--- a/frontend/app/components/magic-link-modal/component.js
+++ b/frontend/app/components/magic-link-modal/component.js
@@ -4,6 +4,8 @@ import Component from "@glimmer/component";
 import { tracked } from "@glimmer/tracking";
 
 export default class MagicLinkModal extends Component {
+  @tracked customer;
+  @tracked project;
   @tracked task;
   @tracked duration;
   @tracked comment;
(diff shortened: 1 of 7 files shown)


derrabauke:
> buggy (after saving a report created by a magic link that didn't have a task, it gets set to blank in the frontend)

*ihmo* this is a feature not a bug, but actually bad UX, as the magic-link dialog should not allow you to create links without a selected task. In the end you can not save reports without selecting a task. So it doesn't make sense to me to enable this via magic links? What's the use-case?

c0rydoras:
the task selection gets set to all blank, but the report gets safe

derrabauke:
Is there an issue to that or do you want to target the issues mentioned in https://github.com/adfinis/timed/issues/79?

> the task selection gets set to all blank, but the report gets saved

Either I don't understand your scenario or can't reproduce.

Why not disable the `Copy link` button as long as no task was selected?

c0rydoras:
http://localhost:4200/reports?comment=work%20on%20example&project=2&duration=PT45M&review=true
image

then i add a task
image

then i press save

c0rydoras:
> Why not disable the Copy link button as long as no task was selected?

i thought about that, but originally i wanted to add magic links for activities (see #79), and those can exist without a task

c0rydoras:
i also thought activities would let you save customer/project but thats gone once you refresh :sweat_smile: 

image

c0rydoras:
but yeah - this doesn't really make much sense; #418 does the disabling stuff

PR #419, opened 2024-08-14
fix(frontend/navbar): weird transition

c0rydoras:
image

steps to reproduce:
1. open timed
2. zoom in / shrink window
3. see that weird thing

image

steps to reproduce:
1. open timed
2. zoom in / shrink window
3. expand nav
4. return to normal size
5. navbar looks scuffed

this also slightly improves the mobile nav 

before:
image

after:
image


derrabauke (approved):
for the mobiles :fist_raised:

PR #432, opened 2024-08-20
refactor optimized-power-select

c0rydoras:
replace event listeners from `did-insert` with `on` modifier, also drop weird legacy stuff


frontend/app/components/optimized-power-select/options/component.js
```diff
@@ -1,79 +1,53 @@
...
-          element.removeEventListener("touchmove", touchMoveHandler);
-        }
-      };
-      // Add touch event handlers to detect taps
-      element.addEventListener("touchstart", () => {
-        element.addEventListener("touchmove", touchMoveHandler);
-      });
-      element.addEventListener("touchend", (e) => {
+      },
+      touchend: () => {
```
derrabauke, line 50: Wouldn't need `this.trackTouchMove` also reset in here?

c0rydoras: AFAICT it didn't do that in the original

derrabauke (approved):
Nice refactoring!

PR #444, opened 2024-09-03
chore(backend): test that midnight is a valid to_time for attendances

c0rydoras:


backend/timed/tracking/tests/test_attendance.py
```diff
@@ -127,3 +127,24 @@ def test_attendance_to_before_from(internal_employee_client, attendance_factory)
     assert json["errors"][0]["detail"] == (
         "An attendance may not end before it starts."
     )
+
+
+def test_attendance_can_end_at_midnight(internal_employee_client, attendance_factory):
```
winged, line 132: Might be worth it to use `@pytest.mark.parametrize` on the test above and test a full set of edge cases. Something like

```python
@pytest.mark.parametrize(
    "attendance__from_time, attendance__to_time, set_from_time, set_to_time expect_result",
    [
        (time(7,30), time(8,30), "7:30", "00:00", status.HTTP_200_OK),
        (time(7,30), time(8,30), "7:00", "00:00", status.HTTP_400_BAD_REQUEST),
        ...
    ]
)
def test_attendance_modification(attendance, set_to_time, set_from_time, internal_employee_client):
    ...
```

backend/timed/tracking/tests/test_attendance.py
```diff
@@ -127,3 +127,24 @@ def test_attendance_to_before_from(internal_employee_client, attendance_factory)
...
+    """Test that 0:00 is a valid to_time."""
+    attendance = attendance_factory(
+        user=internal_employee_client.user, from_time=time(7, 30), to_time=time(8, 30)
+    )
+
+    data = {
+        "data": {
+            "type": "attendances",
+            "id": attendance.id,
+            "attributes": {"to-time": "00:00"},
```
winged, line 142: so from_time = 07:00, to_time = 00:00 - what's the expected duration? Might be worth testing as well (unless it's already tested elsewhere)

winged (commented):
Nice work, but might be useful to parametrize instead of writing a new test case

backend/timed/tracking/tests/test_attendance.py
c0rydoras: the expected duration should be 17h (24 - 7), but AFAICT all the duration things for attendances are done in the frontend

backend/timed/tracking/serializers.py
```diff
@@ -108,8 +108,16 @@ def validate(self, data):
         from_time = data.get("from_time", instance and instance.from_time)
         to_time = data.get("to_time", instance and instance.to_time)
 
-        # validate that to is not before from
-        if to_time is not None and to_time < from_time:
+        if to_time is None:
+            return data
```
winged, line 112: Should we not reject this entirely? an attendance always has start and end, so this should never happen (or, if it happens, something's wrong)

winged (changes requested):

backend/timed/tracking/serializers.py
c0rydoras: totally, not sure how to got there in the first place

winged (approved):
Nice, I like it!

PR #445, opened 2024-09-03
fix(backend/attendances): round from_time and to_time to nearest 15min

c0rydoras:
refactor(backend): move rounding into `timed.utils`
chore(backend): added tests for rounding in `timed.utils`


winged (approved):
LGTM!

PR #446, opened 2024-09-03
fix(backend): allow from_time and to_time to be the same

c0rydoras:
i noticed UX is worse when you can't have the same from and to time, also this would make it so 0:00 to 24:00 attendances aren't possible

### DIFF
--- a/backend/timed/tracking/serializers.py
+++ b/backend/timed/tracking/serializers.py
@@ -109,11 +109,6 @@ class AttendanceSerializer(ModelSerializer):
         from_time = round_time(data.get("from_time", instance and instance.from_time))
         to_time = round_time(data.get("to_time", instance and instance.to_time))
 
-        if to_time == from_time:
-            raise ValidationError(
-                _("An attendance may not start and end at the same time.")
-            )
-
         # allow attendances to end at midnight (00:00)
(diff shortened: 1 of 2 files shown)


winged (changes requested):
Code change LGTM, but commit / PR will need some explanation. Why should this be allowed now?

c0rydoras:
added them to the PR @winged

winged:
Would it make sense to get rid of the ends-at-midnight special case and use 24:00 for that, also in the frontend?

c0rydoras:
the frontend interprets a `to_time` of `0:00` as 24:00

c0rydoras:
so this actually makes it work again

winged:
> the frontend interprets a `to_time` of `0:00` as 24:00

Is this our code or some time library? Cause if it's our code, I'd prefer storing "ends at midnight" as 24:00 instead. Of course we'd need a migration for that, so maybe not worth it..

winged (approved):
Feel free to merge. 

Due to complexity of changing things to be "right", it may not be worth doing the migrations and code changes in the frontend etc.

c0rydoras:
if this is a thing you want implemented feel free to open an issue :)

PR #447, opened 2024-09-03
fix(frontend): fetch projects/tasks when customer/project intersect

c0rydoras:


derrabauke (approved):
Looks legit. Is there a referencing issue?

PR #454, opened 2024-09-04
chore(frontend): fix deprecations by using peekAll instead of fetchAll

c0rydoras:
this works because when prefetching data we already do a `findAll`


derrabauke (approved):
:ok_man:

PR #456, opened 2024-09-04
chore(frontend): set up typescript

c0rydoras:


frontend/app/components/task-selection/component.js
```diff
@@ -56,6 +56,7 @@ export default class TaskSelectionComponent extends Component {
     this._setInitial();
 
     if (this.args.task) {
+      console.log("foo");
```
4censord, line 59: foo

4censord:
is this in prep to move away from ember?

hairmare:
> is this in prep to move away from ember?

:trolleybus:

PR #471, opened 2024-09-17
comment search performance

c0rydoras:
this might make it faster; will have to test more


backend/timed/tracking/migrations/0020_report_search_vector_trigger.py
```diff
@@ -0,0 +1,38 @@
...
+            UPDATE tracking_report SET search_vector = NULL;
+            """,
+            reverse_sql="""
+            DROP TRIGGER IF EXISTS search_vector_trigger
+            ON tracking_report;
+            """,
+        ),
+        migrations.RunPython(
+            compute_search_vector, reverse_code=migrations.RunPython.noop
+        ),
```
winged, line 37: I'd rather have this in Python code.

Move the trigger to either the `Report` model's `save()` method, or into a `pre_save()` signal handler.

While SQL triggers and procedures are appealing, they are an absolute pain to manage and maintain, especially since Django's migration system doesn't know about them.

Peformance-wise this will be slightly worse, but not actually a problem

winged (changes requested):
Awesome work so far!

backend/timed/tracking/migrations/0020_report_search_vector_trigger.py
c0rydoras: i put it in a `post_save` signal handler

backend/timed/tracking/signals.py
```diff
@@ -15,6 +16,12 @@ def update_rejected_on_reports(sender, instance, **kwargs):  # noqa: ARG001
             instance.rejected = False
 
 
+@receiver(post_save, sender=Report)
+def update_search_vector(sender, instance, **kwargs):  # noqa: ARG001
+    """Update comment search vector."""
+    instance.search_vector = SearchVector("comment", config="english")
```
winged, line 22: This will not have any effect: The `post_save` is triggered after the object is saved to DB, and you would need to save it again, causing an infinite loop.

If you just change this to `pre_save`, it will work as intended.

I think it'd be a good idea to write a small test for this:

* Create a report or two
* Trigger the search endpoint and see if it can be found

winged (changes requested):

backend/timed/tracking/signals.py
c0rydoras: setting it to `pre_save` runs into 

```
Failed to insert expression "SearchVector(...)" . F() expressions can only be used to update, not to insert.
```

c0rydoras: > I think it'd be a good idea to write a small test for this

https://github.com/adfinis/timed/blob/c1b5588c975fd404dcd25ab44632efac80eef8bd/backend/timed/tracking/tests/test_report.py#L2099-L2122

such a test already exists :)

winged: `SearchVector(Value(instance.comment))` should work

winged: > such a test already exists :)

Good thing you pointed this out: The filter is still naively doing `comment__search` lookup, completely ignoring the improvements here.

You'll need to update the filter to do something like this:

```python
    def filter_comment(self, queryset, name, value):
        return queryset.filter(
            content_vector=SearchQuery(
                value,
                config="simple", # maybe use "english"?
                search_type="phrase",
            )
        )
```

c0rydoras: it feels a lot faster now

winged (approved):
Nice, I like it!

PR #558, opened 2024-12-23
fix(statistics): use subquery instead of join to avoid cartesian product

### LINKED ISSUE #89: statistic page loading wrong data
some kind of query to the reports endpoint is loading some wrong calculation

ex: https://timed.adfinis.com/statistics?customer=2&fromDate=2023-01-01&ordering=-duration&project=4460&type=task&user=182

image

winged:
The way we filter in the statistics view, any added filter (that affects reports) adds a "dimension" to the cartesian product, exploding the total number of hours reported.

Instead of using JOIN, we do EXISTS(SUBQUERY) now, which should avoid this issue. Might be a tiny bit slower, but let's try to make it correct first, then fast.

Fixes #89


fugal-dy (approved):
I agree with that. The performance bump is relatively high with small numbers but still in an acceptable range in human time. It's very soon outperformed on larger numbers, so I think that's definitely the way to go

winged:
Note this was a one-shot attempt at fixing our issues, created during a hack session with @trowik and @MitanOmar . However it doesn't solve the issue and I think a general refactoring of the whole statistics backend is in order.

winged:
Closing, the resulting work is now in #568

PR #565, opened 2025-01-06
fix(frontend/dark): invert weekend/holiday and workday colours

c0rydoras:
is this really better?

now:
image

after:
image


derrabauke:
> is this really better?
> 

To me it feels more logical to gray out weekends. So, yes. :+1:

derrabauke (approved):

PR #568, opened 2025-01-07
Fix statistics page

winged:
This contains a few commits, but the main one is this:

### fix(statistics): use subquery instead of join to avoid cartesian product

The way we filter in the statistics view, any added filter (that affects
reports) adds a "dimension" to the cartesian product, exploding the total
number of hours reported.

Instead of using JOIN, we do EXISTS(SUBQUERY) now, which should avoid this
issue. Might be a tiny bit slower, but let's try to make it correct first, then fast.

Also fix the catch prefix on the Task statistics filterset - that was wrong
and caused a few incorrect filterings as well.

#### fix(tests): make sure billingtype factory does not crash

Billing Type factory may crash tests because the previous "currency_name" faker
may be rather limited and tries to recreate billing types with the same name,
causing Django to fail (BillingType.name is unique)

#### chore(tests): improve test by using powers-of-two only

This improves the test such that from the resulting sum, we can see
exactly which two reports were affected, because every sum of reports
will now return a distinct sum (any combination of 1,2, and 4 will return
a different result)
This was not the case with 1,2, and 3 before.

#### chore(test): add test to validate multiple-filter behaviour in task stats

Just ensuring we don't run into *that* specific problem anymore

#### fix(dev): correctly run `dbshell` in db container, not a temporary one

The `docker compose run --rm` starts up a new container, which does not
have access to the actual database, so the make target didn't actually work


open-dynaMIX (approved):
Awesome work :rocket:

PR #599, opened 2025-02-10
feat(redmine): Display billable hours

Melkor333:
As we often have to subtract the non-billable hours from project estimation calculations & customer communication, it would be super neat to see the current *billable* hours in the redmine report.

I wondered if it's hard to implement and it looks rather trivial. Though I'm a bit overwhelmed with the testing and would be glad to have some pointers.

### DIFF
--- a/backend/timed/redmine/management/commands/redmine_report.py
+++ b/backend/timed/redmine/management/commands/redmine_report.py
@@ -4,7 +4,7 @@ from datetime import timedelta
 import redminelib
 from django.conf import settings
 from django.core.management.base import BaseCommand
-from django.db.models import Count, Sum
+from django.db.models import Count, Q, Sum
 from django.template.loader import get_template
 from django.utils import timezone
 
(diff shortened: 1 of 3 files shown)


backend/timed/redmine/management/commands/redmine_report.py
```diff
@@ -53,6 +53,11 @@ def handle(self, *args, **options):
             Project.objects.filter(id__in=affected_projects)
             .order_by("name")
             .annotate(total_hours=Sum("tasks__reports__duration"))
+            .annotate(
+                billable_hours=Sum(
+                    "tasks__reports__duration", filter=Q(not_billable=False)
```
Melkor333, line 58: @c0rydoras This `Q` wants to filter the Projects, but I need to filter the `tasks`. Idk the django database model too well. Is there a way to do this "quick & dirty"?
Or should I just make an additional query fort tasks lower down in the `for project...` and filter for project and billability?

I *think* once this works the MR is good to go, unless you're not happy with it only be tested in one of the two tests.

Melkor333:
FWIW I took a second look at the testing today and added the `billable_hours` to both existing report tests. I also made the second test generate some reports with and some without billing to make sure it does a proper calculation.

PR #614, opened 2025-02-19
refactor(frontend): async report ability

c0rydoras:
needed for ember v5.3

### DIFF
--- a/frontend/app/abilities/report.js
+++ b/frontend/app/abilities/report.js
@@ -8,23 +8,49 @@ export default class ReportAbility extends Ability {
     return this.currentUser.user;
   }
 
-  get canEdit() {
-    const isEditable =
-      this.user?.isSuperuser ||
-      (!this.model?.verifiedBy?.get("id") &&
-        (this.model?.user?.get("id") === this.user?.get("id") ||
-          (this.model?.user?.get("supervisors") ?? [])
-            .map((s) => s.id)
-            .includes(this.user?.get("id"))));
-    const isReviewer =
-      (this.model?.taskAssignees ?? [])
-        .concat(
-          this.model?.projectAssignees ?? [],
-          this.model?.customerAssignees ?? [],
-        )
-        .filter((a) => a?.user)
-        .map((a) => a.user.get("id"))
-        .includes(this.user?.get("id")) && !this.model?.verifiedBy?.get("id");
-    return isEditable || isReviewer;
+  get canEditSync() {
+    if (this.user?.isSuperuser) {
+      return true;
+    }
+
+    if (this.model?.verifiedBy?.get("id")) {
+      return false;
+    }
+
+    if (this.model?.user?.get("id") === this.user?.get("id")) {
+      return true;
+    }
+
+    return false;
+  }
+
+  async isReviewer() {
+    return ((await this.model?.taskAssignees) ?? [])
+      .concat(
+        (await this.model?.projectAssignees) ?? [],
+        (await this.model?.customerAssignees) ?? [],
+      )
+      .filter((a) => a?.user)
+      .map((a) => a.user.get("id"))
+      .includes(this.user?.get("id"));
+  }
+
+  async isSupervisee() {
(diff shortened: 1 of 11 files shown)


frontend/app/components/report-row.hbs
```diff
@@ -1,127 +1,131 @@
...
-      @disabled={{not this.editable}}
-      @task={{cs.task}}
-      @on-set-task={{fn this.updateTask cs}}
-      as |t|
+<CanEdit @report={{@report}} as |editable|>
+  {{#let (changeset @report this.ReportValidations) as |cs|}}
+    <ValidatedForm
+      @model={{cs}}
+      class="report-row grid grid-cols-4 gap-2 p-1 lg:grid-cols-[repeat(3,minmax(0,0.9fr)),minmax(0,1.6fr),minmax(0,0.45fr),minmax(2rem,0.6fr),repeat(2,minmax(2rem,0.6fr))] lg:p-1.5 xl:grid-cols-[repeat(3,minmax(0,1.2fr)),minmax(0,1.8fr),minmax(0,0.4fr),minmax(2rem,0.5fr),repeat(2,minmax(2rem,0.5fr))] xl:p-2.5 max-lg:[&>*]:w-full"
+      title={{this.title editable}}
```
derrabauke, line 6:
```suggestion
      title={{if editable (concat "This entry was already verified by " @report.verifiedBy.fullName)}}
```
The `title` via action looks weird. Why not move it into the template :shrug:

frontend/app/helpers/conditional-action.js
derrabauke, line 1: This already exists as [optional](https://github.com/DockYard/ember-composable-helpers?tab=readme-ov-file#optional) in `ember-composable-helpers`.

frontend/app/abilities/report.js
```diff
@@ -8,23 +8,49 @@ export default class ReportAbility extends Ability {
...
+      .includes(this.user?.get("id"));
+  }
+
+  async isSupervisee() {
+    return ((await this.model?.user?.get("supervisors")) ?? [])
+      .map((s) => s.id)
+      .includes(this.user?.get("id"));
+  }
+
+  async canEditAsync() {
```
derrabauke, line 44: It's better than before, but at the same time it bothers me, that the logic for the `canEdit` and `canEditAsync` differs so much, that the naming is a little fallacious. What's the real purpose here? `canReview`?

derrabauke (changes requested):

frontend/app/abilities/report.js
c0rydoras: `canEditSync` -> checks stuff that can be done in sync js
`canEditAsync` -> when `canEditSync` returns false, this checks things that can only be done in async js

frontend/app/helpers/conditional-action.js
c0rydoras: `optional` doesn't do the same thing tho?

derrabauke: But why would you need that? `{{on "click" if this.condition this.myAction noop }}` works fine, doesn't it? Not sure if we could even leave out the [`noop`](https://github.com/DockYard/ember-composable-helpers?tab=readme-ov-file#noop)

frontend/app/abilities/report.js
derrabauke: Pointing out the obvious. :sweat_smile:  -- Seems like we won't come up with a better naming for this check, so let's move on.

frontend/app/helpers/conditional-action.js
c0rydoras: I didn't know `noop` existed :smile:

frontend/app/components/report-row.hbs
c0rydoras: done

frontend/app/helpers/conditional-action.js
c0rydoras: changed

derrabauke (approved):
:rocket:

PR #668, opened 2025-03-24
fix(backend): only allow superusers to edit tasks of archived projects

### LINKED ISSUE #513: [BUG]: I can edit tasks in an archived project
### Description

When i go to the projects view i can see several archived and when i click on them i can change these tasks by e.g. setting a new estimated or refernce.

### Steps to Reproduce

1. go to projects
2. select an archived project
3. edit task
4. press save
5. record is updated

### Expected Behavior

When i choose an archived project then i should not be able to edit it and all the options should be grayed out.

### Attachments

THis is the form that should be disabled on inactive projects

image

### Additional Context

_No response_

### Version

v4.6.0

### Required

MitanOmar:
fix #513


winged (dismissed):
LGTM!

backend/timed/permissions.py
```diff
@@ -65,6 +65,13 @@ def has_object_permission(self, request, view, _obj):
         return self.has_permission(request, view)
 
 
+class IsProjectActive(BasePermission):
+    """Allow Only the project that is not archived."""
```
trowik, line 69:
```suggestion
    """Allows access to object only if a project is not archived."""
```

backend/timed/projects/views.py
```diff
@@ -177,6 +178,8 @@ class TaskViewSet(ModelViewSet):
         (
             # superuser may edit all tasks
             IsSuperUser
+            # only superuser can update tasks of archived projects
+            | (IsUpdateOnly & IsProjectActive)
             # managers may edit all tasks
             | IsManager
```
trowik, line 184: The order is wrong!
This means that anyone who isn't a superuser can update a task if the project is active.

trowik (changes requested):
Please add a test for that!

backend/timed/projects/views.py
winged: Right - and wrong. It's not the order, but the full expression for the permission

```suggestion
            # only superuser can update tasks of archived projects
            | (IsUpdateOnly & IsProjectActive & IsManager)
```

MitanOmar:
the backend tests is failing not because of this PR, and the fix will be #669

c0rydoras:
I think we may need the same thing for archived customers as well maybe?

PR #677, opened 2025-04-01
chore(backend): fix flaky tests

c0rydoras:
fix flaky tests (pytest tried to write randomly seed into somewhere where it couldn't, hence stuff was random)
fix test that broke, at some point


trowik (approved):
Very nice!

Issue #468, opened 2024-09-17
[FEATURE]: worktime balance in navbar


gianklug:
### Is your feature request related to a problem? Please describe.

To check their current worktime balance, a user always has to open the profile page. It would be nice to have it directly visible.

### Describe the solution you'd like

An additional navbar entry showing the current worktime balance. Example:

![image](https://github.com/user-attachments/assets/de1ab9c8-813d-498f-8391-56548be83547)

### Describe alternatives you've considered

Maybe this isn't as good in terms of privacy; e.g. other people looking at one's screen with an open timed window. 

An alternative would be to add a popover when hovering on the user icon; this would still eliminate the need to open an additional page.

### Required

- [X] I have checked past issues to see if this feature has already been requested.

derrabauke:
*imho* the top right bar is quite full already (Reviews, Rejected Reports, etc). But we could actually "hide" that feature in a *obvious* way. For example within the clock. 

- **3h hours minus**
![3h_minus](https://github.com/user-attachments/assets/e743fe71-d866-4335-97fe-c7e660ec7c4e)

- **6h plus**
![6h_plus](https://github.com/user-attachments/assets/d20ab88f-018e-44ac-9d57-669ea213a72a)

Everything above or below 12h is *capped*, but could eventually influence the saturation of the colored background. Colors must adhere to the theme - I just picked random colors for a quick sketch.

derrabauke:
And giving the user a way to turn that *on/off* would be nice as well.

gianklug:
Fair point; I don't usually have reviews or rejected reports in my position; didn't think of that.

I think making the end-user experience more customizable would be great in general, e.g. changing to a dark theme (in the future), setting the default view, etc.

The clock approach looks nice, although with 0 overtime hours, it'll be red for most of the time. 


Maybe something like this would also be a viable option :thinking: 
![image](https://github.com/user-attachments/assets/6740fd0e-6b87-453b-a5cf-b4d9d58f6e5f)

derrabauke:
I've spend some personal free time to implement this. 

Here are my proposals, what you say @gianklug @trowik @c0rydoras?

**indebted**
![Image](https://github.com/user-attachments/assets/05044b25-fb85-48c9-8e6c-222927b1dc4b)

**quite neutral**
![Image](https://github.com/user-attachments/assets/fa094b6d-fd68-4342-9863-a32bb4aa781e)

**lots of overtime**
[...]

PR #744, opened 2025-05-28
feat(timed-clock): dye clock blackground relative to overtime credit

derrabauke:
fix https://github.com/adfinis/timed/issues/468  

The clock will become an indicator for negative and positive worktime and adapts it's color accordingly. Further it will show overtime on hover.


derrabauke:
Made the colored outer part a little thinner to make the general impression a little lighter.

image

c0rydoras:
Could you make this opt-in / toggle-able? (e.g. via some keyboard shortcuts)

derrabauke:
> Could you make this opt-in / toggle-able? (e.g. via some keyboard shortcuts)

Yes, wanted to add a settings view for UI customization as the settings add up over time and key combos are nice for devs but not for non-techies. `will add`

frontend/app/components/timed-clock.js
```diff
@@ -1,15 +1,45 @@
...
 import { scheduleTask } from "ember-lifeline";
 import moment from "moment";
 
+import { COLOR_SCHEME_LIGHT } from "timed/services/appearance";
+
+// local storage key
+export const OVERTIME_FEEDBACK_KEY = "timed-clock-overtime-feedback";
+// color constants
+const MAX_GREEN = 150;
+const MAX_RED = 150;
```
c0rydoras, line 16: personally, i'd prefer sth like a `color-mix` with the colours from the existing themes (via `var()`) and then doing percentages based off of the overtime, that way the colours should always match the used theme

c0rydoras (changes requested):
rebase and maybe address the other comment, otherwise lgtm

c0rydoras (approved):
looks good :chart_with_upwards_trend: :rocket:

PR #795, opened 2025-07-15
fix(backend): use django v5.2 (lts)

c0rydoras:


c0rydoras:
> LGTM, but this is not a fix (MR/commit title)
would be a chore, but to make testing and everything else easier, this should trigger a new releases :) (which fix does)

winged (approved):

PR #841, opened 2025-08-08
frontend: ember 6.4

c0rydoras:
~~currently broken because of issues with ember-responsive~~ ingested into timed repo
~~currently broken because of ember-can~~ hacked around via https://github.com/c0rydoras/ember-can/commit/c95e863484a6c4c202e734553054597d7db9d5a1

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -55,7 +55,7 @@
     "downloadjs": "1.4.7",
     "ember-auto-import": "~2.11.2",
     "ember-basic-dropdown": "^8.2.0",
-    "ember-can": "git+https://github.com/c0rydoras/ember-can#c95e863484a6c4c202e734553054597d7db9d5a1&path:/ember-can",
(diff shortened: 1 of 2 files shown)


derrabauke:
Don't you think it's worth forking the `ember-responsive` package under our namespace? We use it in other projects as well :man_shrugging: 
The [transfer of the project to ember adopted addons](https://github.com/freshbooks/ember-responsive/issues/795) doesn't seem to proceed... 

Another dev also forked it and made it [ember v6 compatible in his fork](https://github.com/freshbooks/ember-responsive/compare/main...felixbuenemann:ember-responsive:ember-v6-compat)

c0rydoras:
regarding `ember-can`, i copied those changes from the [upstream](https://github.com/minutebase/ember-can/commit/8feb0c2805702903d8446c1aa620bda9d946e0fc) and applied them to https://github.com/minutebase/ember-can/pull/182 (which we need for ember data > 4.12)

c0rydoras:
It would make sense to move the fork into the @adfinis namespace

PR #956, opened 2025-11-05
chore(ci/chart): use newer version of yamale

c0rydoras:


derrabauke (dismissed):
Maybe add a note, that if released upstream we can remove it again (v6 seems to be default in the action `main`, but not released yet)

Issue #1047, opened 2026-01-28
[ENHANCEMENT]: Comment Box & History Record for Rescheduling Function


Jana90JL:
### Your proposed enhancement

If a reviewer decides to rebook a timed entry to a different project, the following new functions should be implemented for better transparency:

- If a timed entry is rebooked to a new project, the reviewer has the option to add a comment as to why they are moving the entry (optional field)
- If a timed entry has been rebooked a “history record” should appear to show from where (client, project & task) and from whom (username) the entry was moved. 
- If a timed entry gets rebooked, the user who is rebooking the entry should not be possible to verify the entry while rescheduling it (the verified checkbox should be greyed out/not accessible if the fields “Customer” and/or “Project” are being changed.

- **To be discussed:** How the history record is being displayed and who should see those history entries - Proposal: Only reviewers will be able to see the history records. History redcord can be displayed as an additional icon added in the row for the rebooked entry that opens a popup on hover, showing the following information:

  - User, who rebooked the entry
  - From which client, project and task the entry was moved
  - When it was rebooked

### Enhancement Impact

See above

### Additional context

Estimated Effort: TBD

This serves as the main parent ticket for the following:
- https://github.com/adfinis/timed/issues/32
- https://github.com/adfinis/timed/issues/27

### Required

- [x] I have checked past issues to ensure that this isn't a duplicate.

paraenggu:
Relates to #91 & #1046

PR #1115, opened 2026-03-16
feat: toggle analysis table column

MitanOmar:
fix: #1083


derrabauke (changes requested):
Generally don't scope only to *analysis*. That's a feature we will apply to other tables with a high probability. So all the `analysis-**` prefixed naming --> re-usable component.

c0rydoras (changes requested):

frontend/app/components/analysis-col-group.hbs
```diff
@@ -0,0 +1,5 @@
+<colgroup>
+  {{#each @columns as |column|}}
+    <col class={{concat "w-[" column.width "%]"}} />
```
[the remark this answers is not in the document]
MitanOmar: Fixed

frontend/app/services/analysis-table-columns.js
```diff
@@ -0,0 +1,106 @@
+import { action } from "@ember/object";
+import Service from "@ember/service";
+import { tracked } from "@glimmer/tracking";
+
+export default class AnalysisTableColumnsService extends Service {
+  localStorageKey = "hidden-analysis-table-column";
+  @tracked hiddenColumns = [];
+  @tracked allTableColumns = [
```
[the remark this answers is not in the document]
MitanOmar: removed

frontend/app/components/analysis-columns-modal.js
```diff
@@ -0,0 +1,46 @@
...
+      );
+      return;
+    }
+    this.localhidden.push(columnLabel);
+  }
+
+  @action
+  save() {
+    this.userSettings.updateHiddenColumns("analysisTable", this.localhidden);
+    this.notify.success("The columns are updated successfully");
```
derrabauke: Not necessary, only notify the user if it fails.

frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,83 @@
...
+
+const USER_SETTINGS_KEY = "user-settings";
+
+export default class UserSettingsService extends Service {
+  tableConfigCache = new TrackedObject({
+    // "example-key": {
+    //   all: [],
+    //   hidden: [],
+    // },
+  });
```
derrabauke: What's that for?

frontend/app/utils/user-settings/appearance.js
```diff
@@ -11,12 +9,13 @@ const THEMES = /** @type {const} */ (["old", "regular"]);
...
+  userSettings;
   @tracked _colorScheme;
   @tracked _theme;
 
-  constructor(...args) {
-    super(...args);
+  constructor(userSettings) {
+    this.userSettings = userSettings;
     this.loadConfiguration();
   }
```
derrabauke: What is that? Why not inject the service as normal? Or extend it?

frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,83 @@
...
+  // helper functions for sub settings
+  load(subServiceKey, defaultValue) {
+    const fullKey = `${USER_SETTINGS_KEY}.${subServiceKey}`;
+    const localStorageValue = localStorage.getItem(fullKey);
+    if (!localStorageValue && typeof defaultValue !== "undefined") {
+      localStorage.setItem(fullKey, JSON.stringify(defaultValue));
+      return defaultValue;
+    }
+    return localStorageValue ? JSON.parse(localStorageValue) : defaultValue;
+  }
```
derrabauke: Drop the whole sub-service thingy. Make it flat & simple.

derrabauke (changes requested):

frontend/app/analysis/index/controller.js
```diff
@@ -347,6 +352,80 @@ export default class AnalysisController extends QPController {
...
   });
 
+  prepareTableColumns() {
+    const allTableColumns = [
+      {
+        label: "User",
+        sortable: true,
+        widthClass: "w-[7%]",
+        by: "user__username",
+      },
```
derrabauke: That config belongs into a separate config file. e.g. /frontend/app/configs/analysis-table.js

frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,89 @@
...
+  getTableColumns(tableId) {
+    const defaults = UserSettingsService.TABLE_MAP[tableId] || [];
+    const tableOverrides = this._overrides[tableId] || {};
+
+    return defaults.map((col) => {
+      const userPreference = tableOverrides[col.id];
+      return {
+        ...col,
+        isVisible:
+          userPreference !== undefined ? userPreference : col.isVisible,
```
derrabauke:
```suggestion
          userPreference !== undefined ? userPreference.isVisible : col.isVisible,
```

frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @action
+  updateColumnVisibility(tableId, columnId, isVisible) {
+    // Update the tracked object reference to trigger reactivity
+    this._overrides = {
+      ...this._overrides,
+      [tableId]: {
+        ...(this._overrides[tableId] || {}),
+        [columnId]: isVisible,
```
derrabauke:
```suggestion
        [columnId]: { isVisible },
```

frontend/app/components/analysis-columns-modal.js
```diff
@@ -0,0 +1,59 @@
...
+          this.localhidden.includes(column.label),
+        );
+      }
+      this.args.onClose();
+    } catch (error) {
+      console.log({ error });
+      this.notify.error("An error when updating the columns visibility");
+    }
+  }
+}
```
derrabauke:
```suggestion
import { tracked } from "tracked-built-ins";

export default class AnalysisColumnsModal extends Component {
  @service userSettings;
  @service notify;

  columns = tracked(this.userSettings.getTableColumns("analysis")); // that's all you need

  @action
  toggleColumn(column) {
    column.isVisible = !column.isVisible
    this.columns = [...this.columns]
  }

  @action
  save() {
    try {
      for (const column of this.columns) {
        this.userSettings.updateColumnVisibility(
          "analysis",
          column.label,
          column.isVisible,
        );
      }
      this.args.onClose();
    } catch (error) {
      console.log({ error });
      this.notify.error("An error when updating the columns visibility");
    }
  }
}

```

derrabauke (changes requested):

c0rydoras:
Isn't this blocked by #1047, respectively #1141 (there **will** be conflicts), I thought we discussed that? @MitanOmar @Jana90JL

frontend/app/analysis/index/template.hbs
```diff
@@ -248,85 +249,26 @@
...
-                <col class="w-[10%]" />
-                <col class="w-[10%]" />
-                <col class="w-[21%]" />
-                <col class="w-[8%]" />
-                <col class="w-[5%]" />
-                <col class="w-[5%]" />
-                <col class="w-[5%]" />
-                <col class="w-[5%]" />
-              </colgroup>
+              <AnalysisColGroup @columns={{this.tableColumns}} />
```
derrabauke:
```suggestion
              <Table::ColGroup @columns={{this.tableColumns}} />
```

frontend/app/components/table/col-group.hbs
derrabauke, line 1: Move & rename: `frontend/app/components/table/col-group.hbs`

frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,158 @@
...
+      const data = this.load(key);
+      if (data) {
+        try {
+          this._overrides[key] = data;
+        } catch (e) {
+          console.error("UserSettings: Could not parse localStorage data", e);
+        }
+      }
+    }
+  }
```
derrabauke:
```suggestion
  loadTableConfigurations() {
    const tables = Object.keys(UserSettingsService.TABLE_MAP);
    for (const table of tables) {
      const data = this.load(`${TABLE_CONFIGURATION_PREFIX}.${table}`);
      if (data) {
        try {
          this._overrides[table] = data;
        } catch (e) {
          console.error("UserSettings: Could not load table configurations.", e);
        }
      }
    }
  }
```

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,158 @@
...
+
+  static TABLE_MAP = {
+    analysis: config.APP.analysisTable,
+  };
+
+  constructor(...args) {
+    super(...args);
+    this.loadConfiguration();
+    this._loadSettings();
+  }
```
derrabauke, line 19:
```suggestion
  constructor(...args) {
    super(...args);
    this.loadAppearanceConfigurations();
    this.loadTableConfigurations();
  }
```

frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,158 @@
...
+
+  _persist() {
+    const keys = Object.keys(UserSettingsService.TABLE_MAP);
+    for (const key of keys) {
+      this.save(key, this._overrides[key]);
+    }
+  }
+
+  loadConfiguration() {
```
derrabauke:
```suggestion
  loadAppearanceConfigurations() {
```

frontend/app/components/timed-clock.js
```diff
@@ -49,7 +49,7 @@ export default class TimedClock extends Component {
 
   constructor(...args) {
     super(...args);
-
+    // this.userSettings.of("appearance");
```
derrabauke: Leftover?

frontend/app/components/table-columns-modal.js
```diff
@@ -0,0 +1,33 @@
+import { action } from "@ember/object";
+import { service } from "@ember/service";
+import Component from "@glimmer/component";
+import { tracked } from "tracked-built-ins";
+
+export default class AnalysisColumnsModal extends Component {
+  @service userSettings;
+  @service notify;
+
+  columns = tracked(this.userSettings.getTableColumns("analysis"));
```
derrabauke: The table name should be an argument to the component.

frontend/app/components/table-columns-modal.js
```diff
@@ -0,0 +1,33 @@
...
+    column.isVisible = !column.isVisible;
+    this.columns = [...this.columns];
+  }
+
+  @action
+  save() {
+    try {
+      for (const column of this.columns) {
+        this.userSettings.updateColumnVisibility(
+          "analysis",
```
derrabauke: Same as above

frontend/app/components/table-columns-modal.js
```diff
@@ -0,0 +1,33 @@
...
+          column.label,
+          column.isVisible,
+        );
+      }
+      this.args.onClose();
+    } catch {
+      this.notify.error("An error when updating the columns visibility");
```
derrabauke:
```suggestion
      this.notify.error("Could not update the table column configuration.");
```

frontend/app/components/table/column-picker.hbs
derrabauke, line 1: Unify `frontend/app/components/table-columns-btn.hbs` + `frontend/app/components/table-columns-modal.hbs` to one component (e.g. `frontend/app/components/column-picker.hbs` and adhere to the Data Down, Actions Up (DDAU) architecture.

* data-down = table columns configuration from controller (analysis/index) to component (ColumnPicker)
* action-up: provide a `toggleColumn` handler from the manager (controller) to the component (ColumnPicker)

The analysis/index/controller.js is our *manager* retrieving  the table columns configuration from `userSettingsService`. So it should also manage the updates of said configuration. That way we can keep the *ColumnPicker* **pure**.

ember/frontend/config/environment.js
```diff
@@ -50,6 +51,7 @@ module.exports = function (environment) {
       ],
       EXPORT_LIMIT: 100000,
       OVERTIME_SOFT_LIMIT: process.env.OVERTIME_SOFT_LIMIT || 20,
+      analysisTable,
```
derrabauke: Shouldn't be necessary.

derrabauke (changes requested):
Already a lot better, but still some points to refactor. **keep it up** 🚀

ember/frontend/config/environment.js
MitanOmar: @derrabauke i was trying to include it from the config file, but i was not able to include something from the config without `timed/config/environment`

frontend/app/components/table-columns-btn.js
```diff
@@ -0,0 +1,38 @@
+import { action } from "@ember/object";
+import { service } from "@ember/service";
+import Component from "@glimmer/component";
+import { tracked } from "tracked-built-ins";
+
+export default class AnalysisColumnsBtn extends Component {
```
derrabauke:
```suggestion
export default class ColumnPicker extends Component {
```

frontend/app/components/table/column-picker.hbs
derrabauke: @MitanOmar you've unified the two components which is nice ✨ 

Can we now rename the component? It thought about it again, and as it's highly coupled to the table component, I think it's best to move it to `components/table/column-picker.hbs`/`components/table/column-picker.js`.

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,158 @@
...
+        [columnId]: { isVisible },
+      },
+    };
+    this._persist();
+  }
+
+  loadTableConfigurations() {
+    const keys = Object.keys(UserSettingsService.TABLE_MAP);
+    for (const key of keys) {
+      const data = this.load(key);
```
derrabauke: The signature of `this.load` is `load(subServiceKey, defaultValue)` <-- So you are missing the `subServiceKey` here.

Further, you have some call sites where the `subServiceKey` is set, but it's not actually the `subService` but the table name. Might clear up that confusion.

frontend/app/analysis/index/template.hbs
```diff
@@ -216,6 +216,7 @@
         {{#if this.data.lastSuccessful.value.length}}
           {{#let this.data.lastSuccessful.value as |reports|}}
             <div class="flex justify-end gap-x-2">
+              <TableColumnsBtn @tableName="analysis" />
```
derrabauke:
```suggestion
              <Table::ColumnPicker @tableName="analysis" />
```

derrabauke (changes requested):

frontend/app/analysis/index/template.hbs
MitanOmar: implemented

ember/frontend/app/services/user-settings.js
MitanOmar: @derrabauke i refactor the variable to scope, so i think that it will be more clear now.
what do you think ?

frontend/app/components/table/column-picker.hbs
MitanOmar: @derrabauke implemented

frontend/app/components/table-columns-modal.js
MitanOmar: @derrabauke implemented

frontend/app/components/table-columns-modal.js
MitanOmar: @derrabauke implemented

frontend/app/components/table-columns-modal.js
MitanOmar: @derrabauke implemented

ember/frontend/app/services/user-settings.js
derrabauke: Better. But now if we store table configurations it's using the key `user-settings.analysis`. Would be nicer if we use a subkey for tables like `user-settings.table.analysis`

derrabauke (changes requested):
Almost there. But testing it locally it seems that the layout of the table does not work properly 

* table header & columns are out of sync
Image

frontend/app/services/user-settings.js
derrabauke: @MitanOmar that's what I mean with my last review comment (local storage key)

derrabauke:
> Almost there. But testing it locally it seems that the layout of the table does not work properly
> 
> * table header & columns are out of sync

:point_up_2: that's caused by your rebase in which you removed your `{{#each}}` loop for the table rows visible columns.

Another thing: The `Total:         16:00 h` duration summary in the footer doesn't line up with the duration anymore if the constellation of columns changes. We need to think about a possible solution for that.
image

frontend/app/components/table/column-picker.hbs
```diff
@@ -0,0 +1,38 @@
...
+{{#if this.isModalVisible}}
+  <Modal
+    @visible={{this.isModalVisible}}
+    @onClose={{toggle "isModalVisible" this}}
+    class="md:w-auto"
+    as |modal|
+  >
+    <div class="sm:min-w-[32rem]">
+      <modal.header>
+        <h3>Hide / Show Table Columns</h3>
```
derrabauke:
```suggestion
        <h3>Toggle Table Columns</h3>
```

frontend/app/components/table/column-picker.hbs
```diff
@@ -0,0 +1,38 @@
+<button
+  type="button"
+  title="Toggle Analysis Table Columns Visibility"
```
derrabauke:
```suggestion
  title="Toggle table column visibility"
```

derrabauke (changes requested):

ember/frontend/config/environment.js
derrabauke: That's because you put it into the wrong config file. 
* I was talking about `timed/app/config/analysis-table.js` -> consume it as standard es6 module via  `import analysisTableConfig from "timed/config/analysis-table";` (change to `export default [...]`)
* you put it in `timed/config/analysis-table.js` and consume it via global app environment configuration.

MitanOmar: @derrabauke implemented

frontend/app/components/table/column-picker.hbs
MitanOmar: @derrabauke implemented

frontend/app/components/table/column-picker.hbs
MitanOmar: @derrabauke implemented

ember/frontend/app/services/user-settings.js
MitanOmar: @derrabauke implemented

MitanOmar:
> Almost there. But testing it locally it seems that the layout of the table does not work properly
> 
> * table header & columns are out of sync
> 
> Image

@derrabauke this is also implemented

ember/frontend/app/analysis/index/template.gjs
```diff
@@ -397,36 +341,51 @@ import media from "timed/helpers/media";
...
+                                  @checked={{report.notBillable}}
+                                  @highlight={{true}}
+                                /></Td>
+                            {{else if (eq column.label "Billed")}}
+                              <Td><Checkmark
+                                  @checked={{report.billed}}
+                                  @highlight={{true}}
+                                /></Td>
+                            {{/if}}
+                          {{/each}}
```
c0rydoras, line 396: personally not a big fan of this

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,158 @@
...
+const USER_SETTINGS_KEY = "user-settings";
+
+const getHtml = () => document.querySelector("html");
+
+const COLOR_SCHEME_KEY = "color-scheme";
+const THEME_KEY = "theme";
+
+const THEMES = /** @type {const} */ (["old", "regular"]);
+const COLOR_SCHEME_LIGHT = "light";
+const COLOR_SCHEME_DARK = "dark";
```
c0rydoras: the entire appearance service now lives in `ui-core`, some of this code is duplicated now...

ember/frontend/config/tailwind.config.js
```diff
@@ -142,5 +142,10 @@ module.exports = {
     { pattern: /noUi-/ },
     { pattern: /shepherd-/ },
     "invalid-feedback",
+    "w-[7%]",
+    "w-[10%]",
+    "w-[21%]",
+    "w-[8%]",
+    "w-[5%]",
```
c0rydoras:
```suggestion
```

ember/frontend/config/analysis-table.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * if you are changing the width classes,
+ * don't forget to update frontend/config/tailwind.config.js
+ * since tailwind is not detecting the classes in config file
+ */
```
c0rydoras: ???

c0rydoras:
why two files?
- `ember/frontend/app/config/analysis-table.js`
- `ember/frontend/config/analysis-table.js`

ember/frontend/config/analysis-table.js
c0rydoras: If tailwind isn't picking up on these (you might need to edit a `css` file & save to trigger it re-searching for classes to bundle, I should really document this properly...) you should only need to edit `content` in the tailwind config

But (imo) this belongs somewhere in `app/`, and should therefore "just work" :thinking:

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,158 @@
...
+  @action
+  updateColumnVisibility(tableId, columnId, isVisible) {
+    this._overrides = {
+      ...this._overrides,
+      [tableId]: {
+        ...(this._overrides[tableId] || {}),
+        [columnId]: { isVisible },
+      },
+    };
+    this._persist();
```
derrabauke: Why would you want to save all tables if only one changed? The `_persist` is obsolete (it's merely curried saveAll at all...)

Just call the `this.save(`table.${tableId}`, this._overrides[tableId]);

ember/frontend/app/services/user-settings.js
derrabauke: ☝️ what Arthur says ➕ 

Please remove the stuff which came originally from the appearance service.

@MitanOmar since the app structure changed, we have to think about where to put this `UserSettings` service and if it should still build upon the `Appearance` service inherited from the UI-Core lib.
My thoughts: UI-Core is a re-usable package of components and utils, but our app is something *"independent"* from that (logically, in realistic terms it's under our control and quite coupled...) UI-Core could change and therefore the API (e.g. appearance service). *ergo* I'd vote for creating a `UserSettings` service for the frontend itself.

ember/frontend/config/analysis-table.js
derrabauke: Yep, this file is a left over. Please remove.

MitanOmar: @derrabauke @c0rydoras implemented

ember/frontend/app/services/user-settings.js
MitanOmar: @derrabauke then, i will remove the code of `Appearance` service from user settings service, and just use the `Appearance` service as before, and keep the user settings service in the Timed forntend side

MitanOmar:
> > Almost there. But testing it locally it seems that the layout of the table does not work properly
> > 
> > * table header & columns are out of sync
> 
> 👆 that's caused by your rebase in which you removed your `{{#each}}` loop for the table rows visible columns.
> 
> Another thing: The `Total: 16:00 h` duration summary in the footer doesn't line up with the duration anymore if the constellation of columns changes. We need to think about a possible solution for that. image

@derrabauke do you agree with making it next to each other in the same line, with just a small space between it ?
image

ember/frontend/app/services/user-settings.js
MitanOmar: @derrabauke implemented, and the tests are moved to ui core

ember/frontend/app/services/user-settings.js
MitanOmar: @derrabauke implemented as well

derrabauke:
> @derrabauke do you agree with making it next to each other in the same line, with just a small space between it ? 

:thinking: ... @MitanOmar ideally we would only show it if the `duration` column is visible and if so, it's underneath the duration column as it was. Do you think you can make this work?

ember/frontend/app/services/user-settings.js
derrabauke: @MitanOmar *nope* -- it's still [there](https://github.com/adfinis/timed/pull/1115/changes#diff-804ff5286d985c383c9e12f2b6c41e2cdc17c809fe21a41e732ccbdbfc890d9cR82)!?

ember/ui-core/tests/unit/services/appearance-test.ts
```diff
@@ -0,0 +1,31 @@
...
+  setupTest(hook);
+
+  test("it exists", function (assert) {
+    const appearanceService = this.owner.lookup("service:appearance");
+    assert.ok(appearanceService);
+  });
+
+  test("it can change theme and scheme", function (assert) {
+    const appearanceService = this.owner.lookup("service:appearance");
+    assert.ok(appearanceService);
```
derrabauke: You already tested this some lines above in `it exists`.

ember/ui-core/tests/unit/services/appearance-test.ts
```diff
@@ -0,0 +1,31 @@
...
+    appearanceService.loadConfiguration();
+    const defaultTheme = "old";
+    assert.strictEqual(defaultTheme, appearanceService.theme);
```
derrabauke: Why assign when you only use it once?

ember/ui-core/tests/unit/services/appearance-test.ts
```diff
@@ -0,0 +1,31 @@
...
+    appearanceService.setTheme("regular");
+    assert.strictEqual(appearanceService.theme, "regular");
+    const colorScheme = ["dark", "light"];
+    // since it's checking prefers color scheme in the browser, then it depend on the test machine
+    assert.strictEqual(
+      colorScheme.includes(appearanceService.colorScheme),
+      true,
+    );
```
derrabauke: Vague test environments are bad. Either get it fixed or leave the test out as it won't provide any value.

derrabauke (changes requested):

ember/ui-core/tests/unit/services/appearance-test.ts
MitanOmar: @derrabauke to clarify that we are testing the default case

ember/ui-core/tests/unit/services/appearance-test.ts
MitanOmar: @derrabauke implemented

ember/ui-core/tests/unit/services/appearance-test.ts
MitanOmar: @derrabauke implemented

MitanOmar:
> > @derrabauke do you agree with making it next to each other in the same line, with just a small space between it ?
> 
> 🤔 ... @MitanOmar ideally we would only show it if the `duration` column is visible and if so, it's underneath the duration column as it was. Do you think you can make this work?

@derrabauke we can not handle it 100%, because since the duration column coming after `User` and `Date`, there is a case that the user choose to hide those both columns, so the duration column is coming to the beginning, and it will never be aligned with the start column since there is word `Total:` before it.

but there is something else, we can hide the duration value of the duration column is hidden, without aligning it 100%

ember/frontend/app/services/user-settings.js
MitanOmar: @derrabauke removed

c0rydoras:
> Another thing: The Total: 16:00 h duration summary in the footer doesn't line up with the duration anymore if the constellation of columns changes. We need to think about a possible solution for that. image

Should be fine if you use the same colgroups in the footer as you do for the header/table :thinking:

ember/frontend/app/components/table/column-picker.gjs
```diff
@@ -0,0 +1,89 @@
...
+      class="btn btn-default"
+      {{on "click" (toggle "isModalVisible" this)}}
+      ...attributes
+    >
+      <FaIcon @icon="gear" @prefix="fas" @size="sm" />
+    </button>
+    {{#if this.isModalVisible}}
+      <Modal
+        @visible={{this.isModalVisible}}
+        @onClose={{toggle "isModalVisible" this}}
```
c0rydoras, line 58: instead of toggle, you could use `fn (mut this.isModalVisible) false` & `fn (mut this.isModalVisible) true`

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,81 @@
...
+  updateColumnVisibility(tableId, columnId, isVisible) {
+    this._overrides = {
+      ...this._overrides,
+      [tableId]: {
+        ...(this._overrides[tableId] || {}),
+        [columnId]: { isVisible },
+      },
+    };
+    this.save(`table.${tableId}`, this._overrides[tableId]);
+  }
```
c0rydoras, line 65: consider making this more readable, e.g:

```suggestion
updateColumnVisibility(tableId, columnId, isVisible) {
  const tableOverrides = this._overrides[tableId] || {};

  this._overrides = {
    ...this._overrides,
    [tableId]: {
      ...tableOverrides,
      [columnId]: { isVisible }
    }
  };

  this.save(`table.${tableId}`, this._overrides[tableId]);
}
```

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,81 @@
...
+      const data = this.load(`table.${key}`);
+      if (data) {
+        try {
+          this._overrides[key] = data;
+        } catch (e) {
+          console.error("UserSettings: Could not parse localStorage data", e);
+        }
+      }
+    }
+  }
```
c0rydoras, line 79: nit: `Object.keys(UserSettingsService.TABLE_MAP).forEach` instead of for loop

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,81 @@
...
+      return {
+        ...col,
+        isVisible:
+          userPreference !== undefined ? userPreference.isVisible : true,
+      };
+    });
+  }
+
+  @action
+  updateColumnVisibility(tableId, columnId, isVisible) {
```
c0rydoras, line 55: what are `tableId`/`columnId` here?

consider adding comment(s), preferably jsdoc with type annotations

ember/frontend/app/analysis/index/template.gjs
c0rydoras: still not a big fan of this, but low prio when compared to the other (non nitpick) comments

ember/frontend/app/services/user-settings.js
MitanOmar: @c0rydoras what is the advantage that we will got with using Object.keys ?

ember/frontend/app/components/table/column-picker.gjs
MitanOmar: @c0rydoras it's working as it is at the moment

ember/frontend/app/services/user-settings.js
c0rydoras: more readable imo

you can also replace the `if (data)` with an early return so we have less nesting

ember/frontend/app/components/table/column-picker.gjs
c0rydoras: Yes, but using `fn` and `mut`:
- doesn't use an external library (ember-composable-helpers, which I'd like to remove at some point)
- is less sloppy (you can't click the button to open it when it's opened, you can't close it if it's not visible, so why are we toggling things?)

ember/frontend/app/components/table/column-picker.gjs
```diff
@@ -0,0 +1,89 @@
...
+    <button
+      type="button"
+      title="Toggle table column visibility"
+      class="btn btn-default"
+      {{on "click" (toggle "isModalVisible" this)}}
+      ...attributes
+    >
+      <FaIcon @icon="gear" @prefix="fas" @size="sm" />
+    </button>
+    {{#if this.isModalVisible}}
```
c0rydoras, line 55: also this is redundant (the modal component does this by itself)

ember/frontend/app/components/table/column-picker.gjs
```diff
@@ -0,0 +1,89 @@
...
+        this.userSettings.updateColumnVisibility(
+          this.tableName,
+          column.label,
+          column.isVisible,
+        );
+      }
+      this.isModalVisible = !this.isModalVisible;
+    } catch {
+      this.notify.error("Could not update the table column configuration.");
+    }
```
c0rydoras, line 43: when would this throw an error? and why

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,81 @@
...
+    return localStorageValue ? JSON.parse(localStorageValue) : defaultValue;
+  }
+
+  save(scope, value) {
+    const fullKey = `${USER_SETTINGS_KEY}.${scope}`;
+    localStorage.setItem(fullKey, JSON.stringify(value));
+  }
+
+  clean(scope) {
+    const fullKey = `${USER_SETTINGS_KEY}.${scope}`;
```
c0rydoras: you have
```js
const fullKey = `${USER_SETTINGS_KEY}.${scope}`;
```

3 times here, consider something like

```js
const keyFor = (scope) => `${USER_SETTINGS_KEY}.${scope}`;

[...]

export default class UserSettingsService extends Service {
[...]
 save(scope, value) {
    localStorage.setItem(keyFor(scope), JSON.stringify(value));
  }
  clean(scope) {
    localStorage.removeItem(keyFor(scope));
  }
[...]
```

ember/frontend/app/components/timed-clock.gjs
```diff
@@ -33,7 +33,7 @@ const overtimeToOpacity = (overtime) => {
 
 export default class TimedClock extends Component {
   @service currentUser;
-  @service appearance;
+  @service userSettings;
```
derrabauke: Can't find any usage of the service in the component. I think it's obsolete here.

ember/ui-core/tests/unit/services/appearance-test.ts
derrabauke: Just remove the whole test as it's now out of context of this PR. We can add in a separate PR if you want.

ember/ui-core/tests/test-helper.ts
derrabauke, line 1: same here -- remove.

ember/frontend/app/analysis/index/template.gjs
derrabauke, line 1: The layout of the colgroups of the head vs. table don't line up properly. As you can see in the screenshot below the comment field for example reaches to far into the next column.

Image

ember/frontend/app/analysis/index/template.gjs
```diff
@@ -277,6 +269,7 @@ const AnalysisIndexTemplate = <template>
           {{#if @controller.data.lastSuccessful.value.length}}
             {{#let @controller.data.lastSuccessful.value as |reports|}}
               <div class="flex justify-end gap-x-2">
+                <ColumnPicker @tableName="analysis" />
```
derrabauke: Can we move the `ColumnPicker` to the end of the button group? Having it in first position will shift it unnecessarily when selecting rows for edit.

Image

derrabauke (changes requested):

ember/frontend/app/analysis/index/template.gjs
MitanOmar: @derrabauke it can be that you ware in another branch, and then switched the branch to `Table Columns` without rerunning the frontend.
this is happen because the CSS classes are not compiled

derrabauke: Ah ok, my fault.

ember/frontend/app/components/timed-clock.gjs
MitanOmar: @derrabauke removed

ember/ui-core/tests/test-helper.ts
MitanOmar: @derrabauke removed

ember/frontend/app/analysis/index/template.gjs
MitanOmar: @derrabauke moved to the end now

ember/frontend/app/services/user-settings.js
```diff
@@ -0,0 +1,81 @@
...
+  static TABLE_MAP = {
+    analysis: analysisTableConfig,
+  };
+
+  constructor(...args) {
+    super(...args);
+    this.loadTableConfigurations();
+  }
+
+  // helper functions for sub settings
```
derrabauke:
```suggestion
```

ember/frontend/tests/unit/services/user-settings-test.js
```diff
@@ -0,0 +1,33 @@
...
+    const analysisTable = userSettingsService.getTableColumns("analysis");
+    // check if analysis table have 12 columns
+    assert.strictEqual(analysisTable.length, 12);
+    // remove one
+    userSettingsService.updateColumnVisibility(
+      "analysis",
+      analysisTable[0].label,
+      false,
+    );
+    // now it should be 11 columns
```
derrabauke:
```suggestion
    assert.strictEqual(analysisTable.length, 12);
    userSettingsService.updateColumnVisibility(
      "analysis",
      analysisTable[0].label,
      false,
    );
```
don't comment the obvious ;-)

derrabauke (changes requested):
tiny nitpick/cleanup

ember/frontend/tests/unit/services/user-settings-test.js
MitanOmar: @derrabauke done

ember/frontend/app/services/user-settings.js
MitanOmar: @derrabauke done

derrabauke (approved):

PR #1116, opened 2026-03-16
feat: Additional Sums for Statistics Tab

MitanOmar:
fix: #1082

todo:
* [x] Writing tests


backend/timed/serializers.py
```diff
@@ -9,11 +9,28 @@ class TotalTimeRootMetaMixin:
...
+            aggregate_list["total_remaining_effort"] = Sum("total_remaining_effort")
+        if "most_recent_remaining_effort" in self.fields:
+            aggregate_list["total_remaining_effort"] = Sum(
+                "most_recent_remaining_effort"
+            )
         if many:
             view = self.context["view"]
             queryset = view.filter_queryset(view.get_queryset())
-            data = queryset.aggregate(total_time=Sum(self.duration_field))
+            data = queryset.aggregate(**aggregate_list)
```
winged: I would prefer multiple .aggregate() calls instead. Let's not do the dictionary indirection if it doesn't have any apparent advantage

winged (commented):
Nice work so far, but I think it is not quite correct just yet

winged (changes requested):
Revised one of my comments, but there's still work to do

winged (changes requested):
I think the code LGTM now. But please squash the commits and write a commit message that describes it. We don't need to have all the changes inside the PR visible in the history

winged (approved):
LGTM now!

trowik (approved):
👍

PR #1139, opened 2026-03-30
feat: use textarea in comment field

### LINKED ISSUE #65: Kommentarfeld mehrzeilig
Mir ist in letzter Zeit aufgefallen, dass das Kommentarfeld beim Rapportieren teilweise immer mehr an Relevanz gewonnen hat.
Es ist nun etwas mühselig auf einer Zeile, wenn man ein wenig mehr schreibt.
Alternativ könnte das Kommentarfeld "einfach" Zeilenumbrüche unterstützen, dann kann man den Text in einem editor erfassen und kopieren.

MitanOmar:
fix: #65


derrabauke (dismissed):
🎉

frontend/app/styles/utilities.css
```diff
@@ -9,3 +9,13 @@
...
+
+.text-truncate-2 {
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  line-clamp: 2;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: pre-line;
+}
```
c0rydoras: ❓

c0rydoras (changes requested):

c0rydoras (changes requested):

frontend/app/styles/base/form.css
```diff
@@ -50,3 +50,36 @@ select,
...
+  --max-height: 3.6em;
+}
+
+.content-line-3 {
+  --max-height: 5.4em;
+}
+
+.content-line-4 {
+  --max-height: 7em;
+}
```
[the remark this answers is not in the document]
MitanOmar: @c0rydoras applying it with tailwind classes will need 
```CSS
h-[2.45em] transition-[height] duration-300 ease-in-out resize-none overflow-hidden focus:h-[var(--max-height,2.45em)] focus:overflow-y-auto
```
to be applied in a few different places.

one simple class here will be more clear for reviewers and other devs to know how are we implementing it.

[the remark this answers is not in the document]
MitanOmar: cc @derrabauke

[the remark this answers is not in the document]
MitanOmar: @c0rydoras i don't know if that is what you mean ?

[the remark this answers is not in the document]
c0rydoras: > `h-[2.45em] transition-[height] duration-300 ease-in-out resize-none overflow-hidden focus:h-[var(--max-height,2.45em)] focus:overflow-y-auto` 

please avoid magic numbers like `[2.45em]`, instead use `h-<number` (https://v3.tailwindcss.com/docs/height), this guarantees that spacing is consistent(ish) throughout the application.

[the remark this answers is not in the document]
c0rydoras: > to be applied in a few different places.

So create a `ReportCommentInput`/`ReportCommentTextarea` component and use that in those places?, something like this:

```gts
const contentLineClass = (comment) => {
  const lines = (comment || "").split("\n").length;
  if (lines <= 1) {
    return "content-line-1";
  }
  if (lines === 2) {
    return "content-line-2";
  }
  return lines > 3 ? "content-line-4" : "content-line-3";
};

const ReportCommentTextarea = <template>
  <textarea {{on "input" @onInput}} value={{@value}} ...attributes class="duration-300 ease-in-out resize-none overflow-hidden focus:overflow-y-auto [...] {{contentLineClass @value}}" />
</template>;
```

frontend/app/helpers/report-comment-length-class.js
```diff
@@ -0,0 +1,20 @@
...
+  const lines = (commentContent || "").split("\n").length;
+  if (lines <= 1) {
+    return "content-line-1";
+  } else if (lines === 2) {
+    return "content-line-2";
+  } else if (lines === 3) {
+    return "content-line-3";
+  }
+  return "content-line-4";
+}
```
c0rydoras:
```js
import Component from "@glimmer/component";

const HEIGHT_MAP = {
  1: "h-10",
  2: "h-16",
  3: "h-24",
};

export default class CommentField extends Component {
  get heightClass() {
    const lines = (this.args.value || "").split("\n").length;
    return HEIGHT_MAP[lines] || "h-32";
  }
}
```

```hbs
<textarea ...attributes value={{@value}} {{on "input" @onInput}} class="[...] {{this.heightClass}} {{if @customerVisible 'customer-comment'}}"></textarea>
```

something like this would be a lot cleaner

frontend/app/helpers/report-comment-length-class.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default helper(reportCommentLengthClass);
```
c0rydoras: this should (imo) just be a regular function in the js side of the compoent its used in (nitpick)

c0rydoras (commented):
I hope this makes a bit more sense

frontend/app/helpers/report-comment-length-class.js
MitanOmar: @c0rydoras 

* instead of implementing this JS getter for each component, we write it as a helper, and we can use it in anywhere we have the text area (tracking bar, report raw, analysis edit page).

* as explained in the [requirements](https://github.com/adfinis/timed/issues/65#issuecomment-4088646409), the text area should `auto expand on focus` and return to the old height, to not break the page layout.

* we found this trick with CSS variable applied on focus only, to ignore complex JS code that is watching the interaction of the text area component

MitanOmar: @c0rydoras i tested the way that you said, and it's not returning to the old height if the focus is left

c0rydoras: > "auto expand on focus"

what do you mean with this?

> instead of implementing this JS getter for each component, we write it as a helper, and we can use it in anywhere we have the text area (tracking bar, report raw, analysis edit page).

having a dedicated comment field component is objectively better.

> we found this trick with CSS variable applied on focus only, to ignore complex JS code that is watching the interaction of the text area component

we? as I do not know what you're trying to achieve I can't really judge this...

>  i tested the way that you said, and it's not returning to the old height if the focus is left

All the code I gave you were examples/approximations, **not** tested implementations that can be copied 1:1.

frontend/app/analysis/edit/template.hbs
```diff
@@ -116,22 +116,10 @@
...
-                        title={{if
-                          model.task.project.customerVisible
-                          "This project's comments are visible to the customer"
-                        }}
-                        {{on "change" (pick "target.value" fi.update)}}
-                        value={{fi.value}}
+                      <ReportCommentTextarea
+                        @comment={{fi.value}}
+                        @customerVisible={{model.task.project.customerVisible}}
+                        @update={{fi.update}}
```
c0rydoras:
```suggestion
                        @onUpdate={{fi.update}}
```

c0rydoras: or `@onChange`

frontend/app/components/report-comment.hbs
```diff
@@ -0,0 +1,20 @@
...
+  type="text"
+  class="form-textarea comment-field rounded
+    {{if @customerVisible 'customer-comment'}}
+    {{this.lengthClass @comment}}"
+  placeholder="Comment"
+  name="comment"
+  id="row-comment"
+  value={{@comment}}
+  disabled={{@disabled}}
+  title="{{if @customerVisible '<Customer-visible comments>
```
c0rydoras: ?

frontend/app/analysis/edit/template.hbs
```diff
@@ -116,22 +116,10 @@
...
-                            'customer-comment'
-                          }}"
-                        title={{if
-                          model.task.project.customerVisible
-                          "This project's comments are visible to the customer"
-                        }}
-                        {{on "change" (pick "target.value" fi.update)}}
-                        value={{fi.value}}
+                      <ReportCommentTextarea
+                        @comment={{fi.value}}
```
c0rydoras:
```suggestion
                        @value={{fi.value}}
```

frontend/app/analysis/edit/template.hbs
```diff
@@ -116,22 +116,10 @@
...
-                            model.task.project.customerVisible
-                            'customer-comment'
-                          }}"
-                        title={{if
-                          model.task.project.customerVisible
-                          "This project's comments are visible to the customer"
-                        }}
-                        {{on "change" (pick "target.value" fi.update)}}
-                        value={{fi.value}}
+                      <ReportCommentTextarea
```
c0rydoras:
```suggestion
                      <ReportComment
```

nitpick: it being a textarea doesn't have to be in the component name :)

frontend/app/styles/base/form.css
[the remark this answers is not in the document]
c0rydoras: `content-line` is still here 🤔

c0rydoras (changes requested):
already looks a lot better! 📈 🚀

frontend/app/components/report-comment.hbs
MitanOmar: @c0rydoras there are two title applied for the title attribute

the first was exist before, which was if the comment is visible for the customer
and the second when the report is reviewed it become disabled, and we are showing the comment content in the title tag as well

MitanOmar:
@c0rydoras i have no way with avoiding the extra classes to control the height change, except with JS code

c0rydoras:
@MitanOmar i described it here https://github.com/adfinis/timed/pull/1139#discussion_r3022695948, `max-h-` and `h-`, my point was not "use JS to do it", but that we don't need/nor want classes for this

frontend/app/components/report-comment.hbs
c0rydoras: yes, but previously it was `"This project's comments are visible to the customer"`, not `<Customer-visible comments>`, what prompted the change?

frontend/app/components/report-comment.js
```diff
@@ -0,0 +1,15 @@
...
+    if (lines <= 1) {
+      return "content-line-1";
+    } else if (lines === 2) {
+      return "content-line-2";
+    } else if (lines === 3) {
+      return "content-line-3";
+    }
+    return "content-line-4";
+  }
+}
```
c0rydoras, line 25: move this into the component

c0rydoras (dismissed):

frontend/app/components/report-comment.js
MitanOmar: @c0rydoras just to save some time, could you please implement it in the way that you see it's correct, i have other tasks that is pending

frontend/app/styles/utilities.css
derrabauke: `removed`

frontend/app/styles/base/form.css
```diff
@@ -50,3 +50,36 @@ select,
...
+
+.report-row .ember-basic-dropdown-trigger,
+.report-row .form-list-cell {
+  align-self: start;
```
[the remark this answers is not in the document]
derrabauke: `done`

frontend/app/styles/base/form.css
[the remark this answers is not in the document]
derrabauke: Tried it with directly including tailwind classes -> didn't work. Proxy via own `content-line-x` classes + `@apply` works fine. :man_shrugging: Good enough for right now I'd say.

frontend/app/helpers/report-comment-length-class.js
derrabauke: `done`

frontend/app/helpers/report-comment-length-class.js
derrabauke: See https://github.com/adfinis/timed/pull/1139#discussion_r3052160564

derrabauke: `removed`

frontend/app/analysis/edit/template.hbs
derrabauke: `done`

frontend/app/components/report-comment.hbs
derrabauke: `done`

frontend/app/analysis/edit/template.hbs
derrabauke: `done`

derrabauke (approved):

frontend/app/styles/base/form.css
```diff
@@ -50,3 +50,34 @@ select,
...
+}
+
+.content-line-4 {
+  @apply focus:h-32;
+}
+
+.report-row .ember-basic-dropdown-trigger,
+.report-row .form-list-cell {
+  @apply self-start;
+}
```
c0rydoras, line 83: :(

derrabauke:
https://github.com/adfinis/timed/pull/1164

Issue #660, opened 2025-03-18
[ENHANCEMENT]: Improve contrast of needs review and non-billable in light mode


4censord:
### Your proposed enhancement

In light mode, the difference between ticked and non-ticked non-billable is minimal.
At least for my eyes, this always takes a few closer looks to determin if its ticked or not

![Image](https://github.com/user-attachments/assets/9aef818c-973f-47a7-945f-7e00677aa2da)

### Enhancement Impact

Better contrast

### Required

- [x] I have checked past issues to ensure that this isn't a duplicate.

c0rydoras:
Could you provide us with an example of "better contrast" (e.g. by playing with the inspect tool)

4censord:
I was thinking of just making the light blue a bit more powerfull:

![Image](https://github.com/user-attachments/assets/c159bf9f-19b3-4b2c-9c23-66f4b6d1eb0f)

or what would also work is changing from "lightblue on white" to something else, e.g orange: 

![Image](https://github.com/user-attachments/assets/22d33f05-651a-491e-af2e-6dc7432044f7)

PR #1143, opened 2026-04-01
fix(frontend): improve contrast of needs review & non-billable

g-kedev:
Addresses Issue: #660 

image

image

### DIFF
--- a/ember/ui-core/src/styles/base.css
+++ b/ember/ui-core/src/styles/base.css
@@ -4,6 +4,7 @@
 :root {
   /* taken from https://adfinis.com/en/about-us/branding/ */
 
+  --toggle-active: 255 160, 0;
   --dark-blue: 28 46 93;
   --dark-moderate-blue: 46 75 152;
   --green: 46 152 123;
--- a/ember/ui-core/tailwind.config.mjs
+++ b/ember/ui-core/tailwind.config.mjs
@@ -52,6 +52,9 @@ export const theme = {
       DEFAULT: borderColor,
     },
     colors: {
+      toggle: {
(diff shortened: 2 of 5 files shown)


c0rydoras:
personally not a big fan of it being yellow, @Jana90JL what would you suggest?

derrabauke:
> personally not a big fan of it being yellow, @Jana90JL what would you suggest?

I also wonder where the color comes from. Will pm you a link to our corporate design.

derrabauke (changes requested):
Could you give us some feedback on the choice of color? Maybe give a alternative as preview? :-)

g-kedev:
Took the Color from the example in the [Issue](https://github.com/adfinis/timed/issues/660) and adjusted it a little bit to make it more visible.

g-kedev:
Changed it to "Sunglow" for now, which is a yellow corporate color that looks like this:
image

Here are some alternatives if yellow/orange isn't good (from the corporate colors): 
- Unclicked for Reference
image

- Sapphire
image

- Sky (which is a journey color for cloud native, so not sure if that would be the correct use) 
image

derrabauke:
> Took the Color from the example in the [Issue](https://github.com/adfinis/timed/issues/660) and adjusted it a little bit to make it more visible.

*sunglow* looks best to me. Could you poste another screen with one toggle on / one off + dark-theme? Cropped to the row is enough.

g-kedev:
> _sunglow_ looks best to me. Could you poste another screen with one toggle on / one off + dark-theme? Cropped to the row is enough.

Sure! 
image
image

derrabauke (approved):

PR #1146, opened 2026-04-01
chore(ci): dependabot cooldown for all ecosystems

### LINKED ISSUE #1145: [ENHANCEMENT]: dependabot cooldown
### Your proposed enhancement

Enable dependabot cooldown to ensure that we don't install zero-days.

### Enhancement Impact

* non-security related changes take longer until they are rolled out
* we only upgrade to dependencies that have been available for a while, the theory is that more review happens on such deps.

### Additional context

Dependabot supports this and it can be added to all ecosystems:
```yaml
    cooldown:
      default-days: 7
```
The value of 7 days seems to be an industry standard that is suugested by zizmor and i also found implementations that use 7 days in the wild (e.g. OpenRewrite).

### Required

- [x] I have checked past issues to ensure that this isn't a duplicate.

hairmare:
Activate a default cooldown of 7 days for all ecosystems.

* fixes #1145

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,8 @@ updates:
       interval: daily
       time: "12:00"
       timezone: "Europe/Zurich"
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: pip
     directory: "/backend"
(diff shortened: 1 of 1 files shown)


c0rydoras (approved):
🔒

PR #1148, opened 2026-04-01
 chore(backend/tracking): ensure attendance date is a workday

c0rydoras:
address #1105

### DIFF
--- a/backend/timed/tracking/factories.py
+++ b/backend/timed/tracking/factories.py
@@ -6,7 +6,10 @@ from random import randint
 from factory import Faker, SubFactory, lazy_attribute
 from factory.django import DjangoModelFactory
 
+from timed.factories import load_providers
 from timed.tracking import models
+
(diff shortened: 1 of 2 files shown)


backend/timed/tracking/factories.py
```diff
@@ -6,8 +6,11 @@
 from factory import Faker, SubFactory, lazy_attribute
 from factory.django import DjangoModelFactory
 
+from timed.factories import load_providers
 from timed.tracking import models
 
+load_providers()
```
c0rydoras, line 12: there has to be a cleaner way to do this

One way I can think of would be to subclass `DjangoModelFactory` to load our custom providers and then use that, but I'm also not quite happy with that.

winged: It's a bit cumbersome but OK IMO

winged (approved):
LGTM

PR #1157, opened 2026-04-16
feat: duration picker updated with mouse wheel

### LINKED ISSUE #36: Durationpicker per Scroll +/- 15 Minuten
Beim Durationpicker wäre es super wenn man auch mit dem Scrollrad 15 / 60 Minuten addieren oder subtrahieren könnte.

MitanOmar:
fix: #36

### DIFF
--- a/frontend/app/components/durationpicker-day.js
+++ b/frontend/app/components/durationpicker-day.js
@@ -1,4 +1,5 @@
 import { action } from "@ember/object";
+import { debounceTask } from "ember-lifeline";
 import moment from "moment";
 
 import DurationpickerComponent from "timed/components/durationpicker";
@@ -25,5 +26,16 @@ export default class DurationpickerDayComponent extends DurationpickerComponent 
       const [h, m] = parseDayTime(value);
       this._change(this._set(h, m));
     }
+  }
+
+  @action
+  wheelAction(event) {
+    event.preventDefault();
+    debounceTask(this, "_wheel", event.deltaY, 100);
+  }
+
+  _wheel(deltaY) {
+    const direction = deltaY > 0 ? -1 : 1;
+    this._addMinutes(this.precision * direction);
   }
 }
--- a/frontend/app/components/durationpicker-day.hbs
+++ b/frontend/app/components/durationpicker-day.hbs
@@ -13,5 +13,6 @@
   title={{@title}}
   {{on "change" this.change}}
   {{on "keyup" this.handleKeyPress}}
+  {{on "wheel" this.wheelAction}}
   {{on "focusout" (optional @onFocusOut)}}
 />


frontend/app/components/durationpicker-day.js
```diff
@@ -26,4 +27,14 @@ export default class DurationpickerDayComponent extends DurationpickerComponent
...
+
+  @action
+  wheel(event) {
+    debounceTask(this, "wheelAction", event.deltaY, 100);
+  }
+
+  wheelAction(deltaY) {
+    const direction = deltaY > 0 ? -1 : 1;
+    this._addMinutes(this.precision * direction);
+  }
```
derrabauke, line 40:
```suggestion
  @action
  wheelAction(event) {
    debounceTask(this, "wheelAction", event.deltaY, 100);
  }

  #wheel(deltaY) {
    const direction = deltaY > 0 ? -1 : 1;
    this._addMinutes(this.precision * direction);
  }
```

Clarify naming.

derrabauke (changes requested):
Nitpick: could you correct the commit message to something which reads like a sentence?

e.g. "feat(duration-picker): adds mouse wheel listener"

*gracias* :-)

frontend/app/components/durationpicker-day.js
MitanOmar: implemented

it's happen that my dev tools was open, and the page height was more then the window height,
so it was scrolling the page  and updating the value.

so by preventing the default, the window scroll will stop if the mouse on duration input

derrabauke: `#` missing. Either that or `_wheel` to show that this is private API.

derrabauke:
Just tested it locally. The *scroll to increase* doesn't feel good. It only increases by 15 even if I scroll really slow, since it's debounced. It would be better to compare the `event.deltaY`/`scrollTop` and increase after a certain distance threshold. Using ~~debounceTask~~ `throttleTask` improves it already massively.

MitanOmar:
@derrabauke i tried that, but then it will go from 15 to 45 directly,  and the user may confuse with the scroll amount and speed.

is that OK ?

frontend/app/components/durationpicker-day.js
```diff
@@ -26,4 +27,15 @@ export default class DurationpickerDayComponent extends DurationpickerComponent
       this._change(this._set(h, m));
     }
   }
+
+  @action
+  wheelAction(event) {
+    event.preventDefault();
+    debounceTask(this, "_wheel", event.deltaY, 100);
+  }
```
derrabauke, line 35:
```suggestion
  @action
  wheelAction(event) {
    event.preventDefault();
    
    this.delta += event.wheelDeltaY;
    if (this.delta > 50 || this.delta < -50) {
      this.wheel(this.delta)
      this.delta = 0;
    }
  }
```

This makes it snappy fast and throttles the scrolling so it's controllable even with the trackpad.

frontend/app/components/durationpicker-day.js
```diff
@@ -1,4 +1,5 @@
 import { action } from "@ember/object";
+import { debounceTask } from "ember-lifeline";
```
derrabauke, line 2: Not necessary.

c0rydoras:
Or, we use the existing durationpicker in the `ui-core` addon that has this already?

https://github.com/adfinis/timed/blob/main/ember/ui-core/src/components/durationpicker.gts#L114-L120

MitanOmar:
@c0rydoras i will close this PR, since we are going to use the implementation inside UI Core

PR #1162, opened 2026-04-16
feat(frontend): begin using `ui-core`

### LINKED ISSUE #721: [ENHANCEMENT]: Sorthand
### Your proposed enhancement

> This is a follow-up for https://github.com/adfinis/timed/issues/501.
> The previous ticket was closed despite being solved only partially.

I would expect the following algorithm for the shorthand feature:

1.  All numerical values are padded to four digits with leading zeros
2.  A colon is inserted after the 2nd digit.

I understand that this is probably because 1 expands to 1:00. However, as 100 is already 1:00, I would expect 15 to be 0:15.

In environments, where only 0, 15, 30 and 45 are allowed for the minute part of the duration, additional shorthands for 1 → 1:00, 2 → 2:00, etc. could be added for all single digit inputs.

### Enhancement Impact

For most cases, the shorthand feature would not change. 
Only in cases, where 15 h are to be reported, people had to input `1500` instead of `15`, which is a much more common time to be reported.

Overall, the shorthand feature would become much more intuitive.

### Additional context

_No response_

### Required

- [x] I have checked past issues to ensure that this isn't a duplicate.

c0rydoras:
The goal of this PR is to start **using** `ui-core` in the main timed frontend.

Note that this is not a full refactor of the timed frontend, the goal is to introduce `ui-core`, show/see if it works, and to incrementally refactor/migrate the rest of the timed frontend in further PRs.

For this purpose, this PR will only use the `ReportDurationpicker` from `ui-core`, as it should []()$^{\text{tm}}$ work as a drop-in placement. (which should fix #721 and also (partially) #36)

## open points

_I'd say these are fine to address in seperate/later PRs_

- [ ] how to ensure `ui-core` is built (while i'm ok with a `prebuild` script, having to do the same for other things such as `test` (with `pretest`) feels rather suboptimal) 
- [ ] merge tailwind config
- [ ] cleanup hoisting mess (see `tsconfig`s, and `.npmrc` in `ui-core`)


.github/workflows/frontend.yaml
```diff
@@ -39,19 +39,24 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
         with:
-          version: 9
+          # install pnpm as specified in `packageManager`
+          package_json_file: "ember/package.json"
+
+      - run: |
+          mkdir -p ~/.local/share/pnpm/store/v10
+          pnpm config set store-dir ~/.local/share/pnpm/store/v10
```
derrabauke, line 47: Is this to match different pnpm versions in the different packages?

ember/ui-core/.npmrc
derrabauke, line 1: Shouldn't we install those at the workspace root level? Hoisting stuff like this looks *ugly*

derrabauke (commented):
Two questions otherwise *LGTM* 🎉

ember/ui-core/.npmrc
c0rydoras: we should, but it doesn't build anymore if we don't hoist these

.github/workflows/frontend.yaml
c0rydoras: see https://github.com/pnpm/action-setup/issues/79, this is also in the [`.github/workflows/ui-core.yaml`](https://github.com/adfinis/timed/blob/3b9f5a2e3ab7aacaffcf07bf1bc58c31201e16f8/.github/workflows/ui-core.yaml#L35-L37), and because we build `ui-core` (which runs type checks), we need it here as well

derrabauke: *wild*

derrabauke (approved):

PR #1169, opened 2026-04-20
Revert "Revert "feat: use textarea in comment field""

derrabauke:
Reverts adfinis/timed#1168

Following changes depend on the re-revert: https://github.com/adfinis/timed/pull/1164


trowik (approved):
let's revert the revert 👌

PR #1174, opened 2026-04-21
docs(README): insert command which adds hosts entry to resolve domain

g-kedev:


trowik (dismissed):
👍

derrabauke (dismissed):

README.md
```diff
@@ -56,6 +56,12 @@ Make sure to set your UID in .env
 echo "UID=$UID" >> .env
 ```
 
+Run the following command to add local hosts entry to resolve domain
+
+```bash
+echo "127.0.0.1 timed.localhost" | sudo tee -a /etc/hosts
```
hairmare, line 62: Shouldn't `*.localhost` always resolve to `127.0.0.1`?

derrabauke: ~~with proper dnsmasq maybe :roll_eyes:~~ yes

hairmare: Are you sure you really don't want to fix your operating system? If you are not using Windows, this is usually a misconfiguration stemming from a funky installation on an os that left open too many opinions.

While the [respective RFC](https://datatracker.ietf.org/doc/html/rfc6761#section-6.3) doesn't strongly mandate this, it's still a misconfig if the following doesn't hold true:

> Users may assume that IPv4 and IPv6 address queries for localhost names will always resolve to the respective IP loopback address

derrabauke: *My* system does resolve it properly without the `/etc/hosts` hack, but sadly not everybody has such a great system. :sweat_smile:  Eventually we should adapt the sentence above to reflect the *optional* character. @g-kedev

hairmare: If there were a bit that documents how to test this before just running a sudo thing, then that would for sure help.

I'm still not convinced that people with a Linux that needs this entry should hack the hosts file... they might have a Linux installation to fix first.

g-kedev: I use an Ubuntu. Don't remember touching any network configs and didn't really find anything concerning this on the fly. Adjusted the README entry a little bit.

README.md
```diff
@@ -56,6 +56,12 @@ Make sure to set your UID in .env
 echo "UID=$UID" >> .env
 ```
 
+In case you get an error regarding the Proxying to https://timed.localhost, when starting Ember outside the container (pnpm start), then you may need to run the following command to add a local hosts entry to resolve the domain
+
+```bash
+echo "127.0.0.1 timed.localhost" | sudo tee -a /etc/hosts
```
c0rydoras, line 62: Haven't read any of the above, but `.localhost` **should** resolve to `127.0.0.1` by itself

c0rydoras:
After doing some research, I came across RFC 6761, which has an entire [section](https://datatracker.ietf.org/doc/html/rfc6761#section-6.3) deciated to this topic.

I think it's fine for us to **not** document workarounds for **questionably** configured systems

c0rydoras:
What would you say? @derrabauke @trowik

derrabauke:
> What would you say? @derrabauke @trowik

Voting also for **not including** the proposed hint. :+1:

c0rydoras:
then we can close this :)

PR #1201, opened 2026-05-11
feat: restore scroll on cancel

g-kedev:
Ref: #60

### DIFF
--- a/ember/frontend/app/analysis/index/controller.js
+++ b/ember/frontend/app/analysis/index/controller.js
@@ -56,6 +56,7 @@ export default class AnalysisController extends QPController {
   @service notify;
   @service abilities;
   @service userSettings;
+  @service scrollRestorer;
 
   @tracked _scrollOffset = 0;
   @tracked _shouldLoadMore = false;
@@ -355,6 +356,7 @@ export default class AnalysisController extends QPController {
 
   @action
   edit(selectedIds = [], event) {
+    this.scrollRestorer.storeScrollPosition();
     const ids = event ? selectedIds : [];
     this.router.transitionTo("analysis.edit", {
       queryParams: {
@@ -377,5 +379,14 @@ export default class AnalysisController extends QPController {
 
   dateFromString(str) {
     return DateTime.fromISO(str);
+  }
+
+  @action
+  async restoreScrollPosition() {
+    await this.loadNext.last;
+    await this.data.last;
+    if (this.scrollRestorer.canRestoreScroll) {
+      this.scrollRestorer.restoreScrollPosition();
+    }
   }
 }
--- a/ember/frontend/app/analysis/index/route.js
+++ b/ember/frontend/app/analysis/index/route.js
@@ -1,7 +1,10 @@
 import Route from "@ember/routing/route";
+import { service } from "@ember/service";
 import { runTask } from "ember-lifeline";
 
 export default class AnalysisIndexRoute extends Route {
+  @service scrollRestorer;
+
   queryParams = {
     rejected: {
       refreshModel: true,
@@ -28,5 +31,8 @@ export default class AnalysisIndexRoute extends Route {
 
   setupController(controller) {
     controller.prefetchData.perform();
+
+    controller.restoreScrollPosition();
+    this.scrollRestorer.storeScrollElement("analysis-scrollable-container");
   }
 }
--- a/ember/frontend/app/services/scroll-restorer.js
+++ b/ember/frontend/app/services/scroll-restorer.js
@@ -1,0 +1,52 @@
+import Service from "@ember/service";
+import { debounceTask } from "ember-lifeline";
+
+export default class ScrollRestorerService extends Service {
+  #canRestoreScroll = false;
+  #scrollTop = null;
+  #scrollElement = null;
+
+  set canRestoreScroll(value) {
+    this.#canRestoreScroll = value;
+  }
+
+  get canRestoreScroll() {
+    return this.#canRestoreScroll && this.#scrollTop !== null;
+  }
+
+  set scrollTop(value) {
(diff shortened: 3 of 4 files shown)


g-kedev:
Todo list:
- ~~generalize service -> Done. Change is in a seperate commit to simplify rollback if generalization of scroll restore service not good.~~
- ~~replace did-insert with router hook or some other better alternative -> extend router setupController hook to restore the scroll position.~~

ember/frontend/app/services/scroll-restorer.js
```diff
@@ -0,0 +1,44 @@
+import Service from "@ember/service";
+import { debounceTask } from "ember-lifeline";
+
+export default class ScrollRestorerService extends Service {
+  canRestoreScroll = false;
+  _scrollTop = null;
+  _scrollElement = null;
```
c0rydoras: In modern javascript, we can use `#` for private variables/functions (which makes them actually private) :)

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements

g-kedev: TIL. Done

g-kedev:
@derrabauke feature works now, thanks! Also adjusted the discussed points, please let me know if the feature lives up to your expectations and if I forgot something. 
P.S.: I put the "base" feature in a commit and the adjustments like "generalization of the service" in a seperate commit, in case of a needed rollback. Will squash them when merging.

ember/frontend/app/analysis/index/controller.js
```diff
@@ -355,6 +356,8 @@ export default class AnalysisController extends QPController {
 
   @action
   edit(selectedIds = [], event) {
+    this.scrollRestorer.storeScrollElement("analysis-scrollable-container");
```
derrabauke: Maybe we can do this once when the controller is getting initialized or is the element re-rendererd at some point? This implies to me, that eventually the use of a custom *modifier* could make sense here, since we are coupling the service to a rendered HTML element and eventually wanna know when it's created and destroyed.

ember/frontend/app/services/scroll-restorer.js
```diff
@@ -0,0 +1,52 @@
...
+  storeScrollPosition() {
+    this.scrollTop = this.scrollElement.scrollTop;
+  }
+
+  restoreScrollPosition() {
+    debounceTask(this, "_restore", this.scrollTop, 200);
+  }
+
+  _restore(height) {
+    this.scrollElement.scrollTo({
```
derrabauke: `this.scrollElement` could be undefined. e.g. selecting a report -> cancel -> edit again -> "Save" -> error

derrabauke (dismissed):

ember/frontend/app/analysis/index/route.js
```diff
@@ -28,5 +28,7 @@ export default class AnalysisIndexRoute extends Route {
 
   setupController(controller) {
     controller.prefetchData.perform();
+
+    controller.send("restoreScrollPosition");
```
derrabauke:
```suggestion
    controller.restoreScrollPosition();
```
`controller.send(...)` is old ember syntax.

ember/frontend/app/services/scroll-restorer.js
g-kedev: Done

ember/frontend/app/analysis/index/route.js
g-kedev: Done

ember/frontend/app/analysis/index/controller.js
g-kedev: Put it in the route (for now). I am not sure if a custom modifier would be overkill, it would be more robust, but is it worth sacrificing simplicity here?

g-kedev: As discussed off-channel with Falk, we will merge it how it is right now and refactor it with a custom modifier, when it will be used also somewhere else.

ember/frontend/app/services/scroll-restorer.js
```diff
@@ -0,0 +1,52 @@
...
+  storeScrollElement(selector) {
+    this.scrollElement = selector;
+  }
+
+  storeScrollPosition() {
+    this.scrollTop = this.scrollElement?.scrollTop;
+  }
+
+  restoreScrollPosition() {
+    debounceTask(this, "_restore", this.scrollTop, 200);
```
c0rydoras, line 42: why the debounce?

c0rydoras (dismissed):
LGTM for a first implementation, maybe add a test :)

c0rydoras:
tests failing + out-of-date :)

c0rydoras:
tests still failing + needs rebase

nitpick: 5e56e51f02d91c84cdbf53a3fecb5c01139a4b48 should be a `chore:` not a `fix:` as it only touches code used for/during testing

ember/frontend/tests/acceptance/analysis-test.js
```diff
@@ -272,4 +272,31 @@ module("Acceptance | analysis", function (hooks) {
...
+
+    const originalScrollTop = scrollableContainer.scrollTop;
+
+    await click("[data-test-edit-selected]");
+    await click("button[data-test-cancel]");
+
+    const currentScrollTop = find("[data-test-reports-table]").scrollTop;
+
+    assert.strictEqual(currentScrollTop, originalScrollTop);
+  });
```
c0rydoras, line 314: maybe add like one or two comments explaining what we're doing/testing

g-kedev: Done. LMK if the comments are as imagined.

trowik (approved):
LGTM

PR #1202, opened 2026-05-11
feat: change needs_review and verified in one save

g-kedev:
Previously you had to first uncheck the needs_review flag, save, check verified and save again to change a report from review to verified. With this change you can do that in one go.

Ref: #60

### DIFF
--- a/backend/timed/tracking/tests/test_report.py
+++ b/backend/timed/tracking/tests/test_report.py
@@ -807,18 +807,35 @@ def test_report_update_by_user(
     assert response.status_code == status.HTTP_403_FORBIDDEN
 
 
+@pytest.mark.parametrize(
+    ("needs_review", "needs_review_flag", "expected_status"),
+    [
+        (True, False, status.HTTP_200_OK),
+        (False, True, status.HTTP_400_BAD_REQUEST),
+    ],
+)
 def test_report_update_verified_and_review_reviewer(
-    internal_employee_client, report_factory, project_assignee_factory
+    internal_employee_client,
+    report_factory,
+    project_assignee_factory,
+    needs_review,
+    expected_status,
+    needs_review_flag,
 ):
     user = internal_employee_client.user
     report = report_factory(duration=timedelta(hours=2))
     project_assignee_factory(user=user, project=report.task.project, is_reviewer=True)
+    user = internal_employee_client.user
+    report = report_factory.create(duration=timedelta(hours=2), review=needs_review)
+    project_assignee_factory.create(
+        user=user, project=report.task.project, is_reviewer=True
+    )
 
     data = {
         "data": {
             "type": "reports",
             "id": report.id,
-            "attributes": {"review": True},
+            "attributes": {"review": needs_review_flag},
             "relationships": {
                 "verified-by": {"data": {"id": user.pk, "type": "users"}}
             },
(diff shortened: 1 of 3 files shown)


g-kedev:
---

---

---

---


Following behavior was implemented:
- Base logic: when clicking either needs_review or verified the other is being disabled
- Additionally in the Bulk edit: when indeterminate state is true on either checkbox, don't disable; when both are checked, don't disabled

backend/timed/tracking/views.py
```diff
@@ -248,6 +249,22 @@ def bulk(self, request):  # noqa: C901
                 _("Only superuser and accountants may bill reports")
             )
 
+        effective_review = (
+            review
+            if review is not None
+            else any(queryset.values_list("review", flat=True))
```
c0rydoras: Is there a particular reason for you doing `any(queryset.values_list("review", flat=True))`?

Iirc `QuerySet.exists` is a thing, why not use that?

winged: I think it *can* make sense if the same queryset is reused and there's not too many results (as it will only query the DB once). However in this instance, I don't think that's the case, as the `.values_list()` call is is explicitly querying the DB IIRC.

If in doubt, measure! Run through a test case and either enable logging to see if this improves query count (and check run time as well!). I can support if you really want to dig into it.

g-kedev: > Is there a particular reason for you doing any(queryset.values_list("review", flat=True))?

No particular reason, just saw that it was previously used like that, thus kept it. Adjusted it, now it uses `.exists()`

> If in doubt, measure! Run through a test case and either enable logging to see if this improves query count (and check run time as well!). I can support if you really want to dig into it.

I'm not entirely sure how to properly measure the performance or query count yet. If you're up for it, I'd really appreciate a quick introduction or walkthrough so I can learn how to do it properly!

backend/timed/tracking/views.py
```diff
@@ -256,11 +271,6 @@ def bulk(self, request):  # noqa: C901
                     _("Reviewer filter needs to be set to verifying user")
                 )
```
c0rydoras: why have this outside of the `if verified is not None:` block?

c0rydoras (changes requested):

backend/timed/tracking/views.py
g-kedev: Hadn't seen it. Adjusted it.

ember/frontend/app/analysis/edit/template.gjs
```diff
@@ -268,7 +273,11 @@ import Void from "timed/components/void";
                             @title={{@controller.toolTipText}}
                             @disabled={{or
                               (not @controller.canVerify)
-                              @controller.needsReview
+                              (and
+                                f.model.review
+                                (notEq f.model.verified null)
+                                (not f.model.verified)
+                              )
```
c0rydoras, line 280: Does this just work? 

_Most of the existing validation for this form is in `app/validations/intersection.js` ^^_

g-kedev: ~~Used to work, apparently it doesn't anymore yes, works (forgot that you have to set the reviewer filter to verify). Would the validation be triggered on every change?~~
Done

backend/timed/tracking/tests/test_report.py
```diff
@@ -810,18 +810,35 @@ def test_report_update_by_user(
     assert response.status_code == status.HTTP_403_FORBIDDEN
 
 
+@pytest.mark.parametrize(
+    ("needs_review", "needs_review_flag", "expected_status"),
```
winged: Rename the params so it's clear which one is used for which


```suggestion
    ("has_needs_review", "set_needs_review", "expected_status"),
```

winged (commented):
Minor nitpick, otherwise LGTM

backend/timed/tracking/tests/test_report.py

PR #1212, opened 2026-05-18
fix(frontend/modal): ensure ember-power-select searchbar can be clicked

MitanOmar:
fix #1133

### DIFF
--- a/ember/frontend/app/components/magic-link-modal.gjs
+++ b/ember/frontend/app/components/magic-link-modal.gjs
@@ -65,6 +65,7 @@ export default class MagicLinkModal extends Component {
       @visible={{@isVisible}}
       @onClose={{@onClose}}
       data-test-magic-link-modal
+      @targetId="ember-basic-dropdown-wormhole"
       class="md:w-auto"
       as |modal|
     >
--- a/ember/frontend/app/components/modal.gjs
+++ b/ember/frontend/app/components/modal.gjs
@@ -28,8 +28,12 @@ const optional = (onClose) => {
   return onClose;
 };
 export default class Modal extends Component {
+  get targetId() {
+    return this.args.targetId ?? TARGET_ID;
+  }
+
   get target() {
-    return document.getElementById(TARGET_ID);
+    return document.getElementById(this.targetId);
   }
 
   <template>
--- a/ember/frontend/tests/acceptance/magic-link-test.js
+++ b/ember/frontend/tests/acceptance/magic-link-test.js
@@ -18,6 +18,23 @@ module("Acceptance | magic links", function (hooks) {
     this.server.createList("report", 5, { userId: user.id });
   });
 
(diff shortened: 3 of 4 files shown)


c0rydoras:
I think it would be cleaner to set `renderInPlace` on the power selects, rather than duct taping it in the modal component

MitanOmar:
> I think it would be cleaner to set `renderInPlace` on the power selects, rather than duct taping it in the modal component

I tried that, but it was not working

c0rydoras:
Could you maybe add a test for this? (It should **fail** without the fix)

c0rydoras:
Regarding `renderInPlace`:

It **is** an option. But to actually use it, you'd first need to add a `@renderInPlace={{@renderInPlace}}`  in `OptimizedPowerSelectComponent` :upside_down_face: 

~~And, you'd also need to revert/change the css hack we introduced in #1158 (otherwise the select is just hidden :zany_face:) ~~ Comment out some css for `ember-basic-dropdown` (not quite sure what this does/why we have it)

```css
/* app/styles/components/basic-dropdown.css */
[...]
/* 
.ember-basic-dropdown-content-wormhole-origin {
  @apply hidden;
}*/
```

ember/frontend/app/components/magic-link-modal.gjs
```diff
@@ -61,6 +61,7 @@ export default class MagicLinkModal extends Component {
       @visible={{@isVisible}}
       @onClose={{@onClose}}
       data-test-magic-link-modal
+      @target-id="ember-basic-dropdown-wormhole"
```
derrabauke:
```suggestion
      @targetId="ember-basic-dropdown-wormhole"
```

component arguments should be in camelCase

MitanOmar: @derrabauke implemented

MitanOmar:
> Could you maybe add a test for this? (It should **fail** without the fix)

@c0rydoras wrote tests that is checking if the component is rendering the right id

c0rydoras:
> > Could you maybe add a test for this? (It should **fail** without the fix)
> 
> @c0rydoras wrote tests that is checking if the component is rendering the right id

@MitanOmar I meant an acceptance test that tests that #1133 is fixed, respectively broken (before the fix).

derrabauke (dismissed):
Tried the `renderInPlace` approach, but that would yield another bunch of issues. Your approach seems to work. Tests look good as well. 👍

c0rydoras (changes requested):
what about the acceptance test (that tests #1133 directly)?

otherwise LGTM

c0rydoras (approved):
acceptance test exists ✅ 
acceptance test fails without the fix ✅ 

_I do have nitpicks, and some opinions (e.g. on passing `id`s around), but this getting fixed is probably more important than bike shedding_

 :shipit:

PR #1227, opened 2026-05-28
Split report by reviewer via modal ( Frontend Side Only )

MitanOmar:
a part of #1121

* [x] split the modal content UI `Original report in the left, and new one in the right`
* [x] rename `new report` to be `Second Report`
* [x] show the label in top of the input
* [x] show the `Task Selection` of the original report
* [x] bring the cancel to `float start`
* [x] split reports for all users who can edit the report
* [x] Automated Tests

### DIFF
--- a/ember/frontend/app/analysis/edit/controller.js
+++ b/ember/frontend/app/analysis/edit/controller.js
@@ -59,6 +59,7 @@ export default class AnalysisEditController extends Controller {
   @service currentUser;
   @service store;
   @service unverifiedReports;
+  @service abilities;
 
   @tracked id;
   @tracked user;
@@ -152,6 +153,10 @@ export default class AnalysisEditController extends Controller {
       return false;
     }
     return this.isReviewer || this.isSuperuser;
+  }
+
+  get canSplitReport() {
+    return this.id?.length === 1;
   }
 
   get canBill() {
--- a/ember/frontend/app/analysis/edit/template.gjs
+++ b/ember/frontend/app/analysis/edit/template.gjs
@@ -15,6 +15,7 @@ import Empty from "timed/components/empty";
 import NotIdenticalWarning from "timed/components/not-identical-warning";
 import PagePermission from "timed/components/page-permission";
 import ReportComment from "timed/components/report-comment";
+import SplitReportModal from "timed/components/split-report-modal";
 import TaskSelection from "timed/components/task-selection";
 import Void from "timed/components/void";
 
@@ -320,6 +321,11 @@ import Void from "timed/components/void";
                             )
                           }}
                         >Reset</button>
+                        <SplitReportModal
+                          @afterSave={{@controller.cancel}}
+                          @disabled={{not @controller.canSplitReport}}
+                          @report-id={{@controller.id}}
+                        />
                       </div>
                       <f.submit />
                     </c.footer>
--- a/ember/frontend/app/components/split-report-modal.gjs
+++ b/ember/frontend/app/components/split-report-modal.gjs
@@ -49,16 +49,6 @@ export default class SplitReportModal extends Component {
     this.isModalVisible = true;
   }
 
-  @action
-  onSetOldTask(task) {
-    this.oldTask = task;
-  }
-
-  @action
-  onSetNewTask(task) {
-    this.newTask = task;
-  }
-
   get remainingOldDuration() {
     const remaining = this.oldDuration.minus(this.newDuration);
     return remaining.as("milliseconds") > 0
@@ -108,17 +98,26 @@ export default class SplitReportModal extends Component {
 
       await this.fetch.fetch(`/api/v1/reports/${reportId}/split`, {
         method: "POST",
-        body: {
-          old_report: {
-            task: this.oldTask.id,
-            comment: this.oldComment,
-            duration: durationTransform.serialize(this.remainingOldDuration),
+        data: {
+          attributes: {
+            updated_original_report: {
+              task: {
+                type: "tasks",
+                id: this.oldTask.id,
+              },
+              comment: this.oldComment,
+              duration: durationTransform.serialize(this.remainingOldDuration),
+            },
+            second_report: {
+              task: {
+                type: "tasks",
+                id: this.newTask.id,
+              },
+              comment: this.newComment,
+              duration: durationTransform.serialize(this.newDuration),
+            },
           },
-          new_report: {
-            task: this.newTask.id,
-            comment: this.newComment,
-            duration: durationTransform.serialize(this.newDuration),
-          },
+          type: "split-reports",
         },
       });
 
@@ -173,7 +172,7 @@ export default class SplitReportModal extends Component {
                 <p class="text-muted-foreground text-sm font-semibold">Original
                   report</p>
                 <TaskSelection
-                  @on-set-task={{this.onSetOldTask}}
+                  @on-set-task={{fn (mut this.oldTask)}}
                   @task={{this.oldTask}}
                   as |t|
                 >
@@ -204,7 +203,7 @@ export default class SplitReportModal extends Component {
               <div class="flex flex-col gap-2">
                 <p class="text-muted-foreground text-sm font-semibold">Second
                   Report</p>
-                <TaskSelection @on-set-task={{this.onSetNewTask}} as |t|>
+                <TaskSelection @on-set-task={{fn (mut this.newTask)}} as |t|>
                   <t.customer @dropdownClass="z-[60]" />
                   <t.project @dropdownClass="z-[60]" />
                   <t.task @dropdownClass="z-[60]" />
@@ -235,6 +234,7 @@ export default class SplitReportModal extends Component {
             <button
               class="btn btn-default"
               type="button"
+              data-test-split-report-cancel
               {{on "click" this.closeModal}}
             >
               Cancel
@@ -242,6 +242,7 @@ export default class SplitReportModal extends Component {
             <button
               class="btn btn-primary ms-auto"
               type="button"
+              data-test-split-report-confirm
               {{on "click" (perform this.splitReport)}}
               disabled={{or (not this.isValidSplit) this.splitReport.isRunning}}
             >
--- a/ember/frontend/app/components/split-report-modal.gjs
+++ b/ember/frontend/app/components/split-report-modal.gjs
@@ -10,7 +10,7 @@ import perform from "ember-concurrency/helpers/perform";
 import { not, or } from "ember-truth-helpers";
 import { Duration } from "luxon";
 
-import DurationpickerDay from "timed/components/durationpicker-day";
+import Durationpicker from "timed/components/durationpicker";
 import Modal from "timed/components/modal";
 import TaskSelection from "timed/components/task-selection";
 import DjangoDurationTransform from "timed/transforms/django-duration";
@@ -194,7 +194,7 @@ export default class SplitReportModal extends Component {
                     aria-label="Comment for original report"
                   />
                 </div>
-                <DurationpickerDay
+                <Durationpicker
                   @value={{this.remainingOldDuration}}
                   @onChange={{this.onRemainingDurationChange}}
                   @title="Remaining duration"
@@ -222,7 +222,7 @@ export default class SplitReportModal extends Component {
                     aria-label="Comment for new report"
                   />
                 </div>
-                <DurationpickerDay
+                <Durationpicker
                   @value={{this.newDuration}}
                   @onChange={{this.onNewDurationChange}}
                   @title="Duration for new report"
--- a/ember/frontend/app/components/split-report-modal.gjs
+++ b/ember/frontend/app/components/split-report-modal.gjs
@@ -1,0 +1,255 @@
+import { fn } from "@ember/helper";
+import { on } from "@ember/modifier";
+import { action } from "@ember/object";
+import { service } from "@ember/service";
+import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
+import pick from "@nullvoxpopuli/ember-composable-helpers/helpers/pick";
+import { task } from "ember-concurrency";
+import perform from "ember-concurrency/helpers/perform";
+import { not, or } from "ember-truth-helpers";
+import { Duration } from "luxon";
+
+import DurationpickerDay from "timed/components/durationpicker-day";
+import Modal from "timed/components/modal";
+import TaskSelection from "timed/components/task-selection";
+import DjangoDurationTransform from "timed/transforms/django-duration";
+
+export default class SplitReportModal extends Component {
+  @tracked isModalVisible = false;
+  @service fetch;
+  @service notify;
+  @service store;
+
+  @tracked newTask = null;
+  @tracked newComment = "";
+  @tracked newDuration = Duration.fromMillis(0);
+
+  @tracked oldTask = null;
+  @tracked oldComment = "";
+  @tracked oldDuration = Duration.fromMillis(0);
+
+  get report() {
+    return this.fetchReport.lastSuccessful?.value;
+  }
+
+  fetchReport = task(async () => {
+    return await this.store.findRecord("report", this.args["report-id"][0]);
+  });
+
+  @action
+  async showModal() {
+    const report = await this.fetchReport.perform();
+    this.oldTask = report.task ?? null;
+    this.oldComment = report.comment ?? "";
+    this.oldDuration = report.duration ?? Duration.fromMillis(0);
+    this.newDuration = Duration.fromMillis(0);
+    this.newComment = "";
+    this.newTask = null;
+    this.isModalVisible = true;
+  }
+
+  @action
+  onSetOldTask(task) {
+    this.oldTask = task;
+  }
+
+  @action
+  onSetNewTask(task) {
+    this.newTask = task;
+  }
+
+  get remainingOldDuration() {
+    const remaining = this.oldDuration.minus(this.newDuration);
+    return remaining.as("milliseconds") > 0
+      ? remaining
+      : Duration.fromMillis(0);
+  }
+
+  @action
+  onNewDurationChange(value) {
+    const totalMs = this.oldDuration.as("milliseconds");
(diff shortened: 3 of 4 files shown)


frontend/app/analysis/edit/controller.js
```diff
@@ -154,6 +154,13 @@ export default class AnalysisEditController extends Controller {
     return this.isReviewer || this.isSuperuser;
   }
 
+  get canSplitReport() {
+    return (
+      this.id.length === 1 &&
```
c0rydoras: `this.id.length === 1` ?

frontend/app/components/modal/overlay.js
```diff
@@ -11,7 +11,7 @@ export default class ModalOverlay extends Component {
 
   @action
   handleClick(e) {
-    if (e.target.id === this.id) {
+    if (e.target.id === this.id && this.args.closeOnBackdropClick !== false) {
```
c0rydoras: what is this for?

frontend/app/components/split-report-modal.js
```diff
@@ -0,0 +1,129 @@
...
+  @service notify;
+  @service store;
+
+  @tracked newTask = null;
+  @tracked newComment = "";
+  @tracked newDuration = moment.duration();
+
+  @tracked oldTask = null;
+  @tracked oldComment = "";
+  @tracked oldDuration = moment.duration();
```
c0rydoras: Is there a reason you didn't use models here?

frontend/app/analysis/edit/controller.js
MitanOmar: This is one of many coming conditions that should be applied before allowing split report,
and the first condition that the selected reports for edit have to be one

frontend/app/components/modal/overlay.js
MitanOmar: this is for disable interactivity with the background, so no one will accidentally close the modal

frontend/app/components/split-report-modal.js
MitanOmar: do you mean  passing a reference from parent to child ( Split Report Component ) ?

frontend/app/analysis/edit/controller.js
c0rydoras: What is `id` here? why are we taking its length? isn't `id` usually an increasing number? (for like records)

frontend/app/components/modal/overlay.js
c0rydoras: What would be the usecase for that?

frontend/app/components/split-report-modal.js
c0rydoras: I meant using ember-data models, for boith?

frontend/app/analysis/edit/controller.js
MitanOmar: image
it's an array of selected reports id as string

frontend/app/components/modal/overlay.js
MitanOmar: user experience preferred after collection opinions from the team

frontend/app/components/split-report-modal.js
MitanOmar: it's not the final version yet, this part was done after a request of making a demo.
and this was easier for quick demo, let's see later what will be the final results

frontend/app/components/modal/overlay.js
c0rydoras: No I mean, of this implementation, why not have the browser handle this for us

frontend/app/analysis/edit/controller.js
c0rydoras: Ahhh, that's cursed, maybe rename it to `ids` or `reports`/`reportIds`? (that would greatly improve readability)

c0rydoras:
you might want to rebase ^^

MitanOmar:
@g-kedev fyi: the modified information is sent to endpoint `/api/v1/reports/{original-report-id}/split`,
and the payload is structured as
```
old_report: {
     task: "original report id",
      comment: "original content or modified content",
       duration: "modified duration",
},
new_report: {
       task: "The new task id"
       comment: "The new content",
       duration: "the new duration",
},
```

in case that you prepared another endpoint, or expected it somewhere else, just comment here, and i will modify the URL in the frontend

ember/frontend/app/components/split-report-modal.gjs
```diff
@@ -0,0 +1,255 @@
...
+  @service notify;
+  @service store;
+
+  @tracked newTask = null;
+  @tracked newComment = "";
+  @tracked newDuration = Duration.fromMillis(0);
+
+  @tracked oldTask = null;
+  @tracked oldComment = "";
+  @tracked oldDuration = Duration.fromMillis(0);
```
c0rydoras, line 30: If we're not going to use ember data models, you could have these as a separate data structure, to improve readability and maintainability :)

```js
class Report {
  @tracked
  task = null;
  
  @tracked 
  comment = "";
  
  @tracked 
  duration = Duration.fromMillis(0);
};


export default class SplitReportModal extends Component {
  report = new Report();
  newReport = new Report();
  [...]
}
```

What's the rationale for not using ember data models?

MitanOmar: because it need deep watching of the changes of the model.
and this was easier for implementation, and also more clear for reviewer

ember/frontend/app/components/split-report-modal.gjs
```diff
@@ -0,0 +1,255 @@
...
+
+  @action
+  onSetOldTask(task) {
+    this.oldTask = task;
+  }
+
+  @action
+  onSetNewTask(task) {
+    this.newTask = task;
+  }
```
c0rydoras: you could inline these in the template with  `fn (mut this.newTask)`, or `fn (mut this.splitReport.task)`

which is what I personally prefer (for readability)

ember/frontend/app/components/split-report-modal.gjs
```diff
@@ -0,0 +1,255 @@
...
+    const ms = Math.min(value?.as("milliseconds") ?? 0, totalMs);
+    this.newDuration = Duration.fromMillis(Math.max(ms, 0));
+  }
+
+  @action
+  onRemainingDurationChange(value) {
+    const totalMs = this.oldDuration.as("milliseconds");
+    const remainingMs = Math.min(value?.as("milliseconds") ?? 0, totalMs);
+    this.newDuration = Duration.fromMillis(Math.max(totalMs - remainingMs, 0));
+  }
```
c0rydoras, line 71: You set `newDuration` in both of these 🤔

Consider adding a comment or two to explain what is happening here.

There's likely a much neater way to implement this, maybe @derrabauke has some ideas?

ember/frontend/app/components/split-report-modal.gjs
```diff
@@ -0,0 +1,255 @@
...
+      return false;
+    }
+    if (
+      !this.oldDuration ||
+      this.newDuration.as("milliseconds") >= this.oldDuration.as("milliseconds")
+    ) {
+      return false;
+    }
+    return true;
+  }
```
c0rydoras, line 85: Have you considered using `ember-changeset-validations`? (see `ember/frontend/app/validators/datetime.js`)

ember/frontend/app/components/split-report-modal.gjs
```diff
@@ -0,0 +1,255 @@
...
+            comment: this.oldComment,
+            duration: durationTransform.serialize(this.remainingOldDuration),
+          },
+          new_report: {
+            task: this.newTask.id,
+            comment: this.newComment,
+            duration: durationTransform.serialize(this.newDuration),
+          },
+        },
+      });
```
c0rydoras, line 122: I would imagine ember data could help us here? Like you could probably create an ember data model for this (not quite sure how well this would work)

Maybe @derrabauke has some input/pointers?

ember/frontend/app/analysis/edit/controller.js
```diff
@@ -154,6 +155,16 @@ export default class AnalysisEditController extends Controller {
     return this.isReviewer || this.isSuperuser;
   }
 
+  get canSplitReport() {
+    return (
+      this.id?.length === 1 &&
+      this.abilities.can(
```
g-kedev: @MitanOmar Currently the reviewer can't split the report, because it uses the canEditSync where the reviewer isn't checked.

MitanOmar: @g-kedev since we already doing a validation if the current user is reviewer before arriving to edit report form, then there is no need for a double check there.
it's working now as expected

MitanOmar:
image

@derrabauke @c0rydoras @Jana90JL @g-kedev @trowik 
should we keep this button disabled, or just remove it
:+1: for removing
:heart: for disabling it

ember/frontend/app/analysis/edit/controller.js
g-kedev: :+1:

ember/frontend/app/analysis/edit/controller.js
```diff
@@ -156,13 +156,7 @@ export default class AnalysisEditController extends Controller {
...
   get canSplitReport() {
-    return (
-      this.id?.length === 1 &&
-      this.abilities.can(
-        "editSync report",
-        this.intersection.lastSuccessful.value.model,
-      )
-    );
+    return this.id?.length === 1;
   }
```
c0rydoras, line 160: > refactor: remove the validation if the current user is a reviewer

AFAICT this just fully removes the validation?

nit: `refactor(frontend): ` instead of `refactor: `

c0rydoras:
instead of `feat: split reports` as commit message, I'd suggest something like `feat(frontend): allow splitting reports`

g-kedev:
> @g-kedev fyi: the modified information is sent to endpoint `/api/v1/reports/{original-report-id}/split`, and the payload is structured as
> 
> ```
> old_report: {
>      task: "original report id",
>       comment: "original content or modified content",
>        duration: "modified duration",
> },
> new_report: {
>        task: "The new task id"
>        comment: "The new content",
>        duration: "the new duration",
> },
> ```
> 
> in case that you prepared another endpoint, or expected it somewhere else, just comment here, and i will modify the URL in the frontend

@MitanOmar  Endpoint is correct. The data structure is a little different, here is an example:
```
{
  "data": {
    "attributes": {
      "new_report": {
        "comment": "some new comment",
        "duration": "00:15:00",
        "task": {
          "type": "tasks",
          "id": 16
        }
      },
      "updated_report": {
        "comment": "some updated comment",
        "duration": "00:15:00",
        "task": {
          "type": "tasks",
          "id": 16
        }
      }
    },
    "type": "report-splits"
  }
}
```
Edit: Will tell you the reworked data structure, once the backend part is finished with the review process (and prob merged)

c0rydoras:
> "duration_as_seconds"

Why not use a `DurationField`?

g-kedev:
> Why not use a `DurationField`?

Since the data is nested, I wasn't sure how to validate the fields via the serializer attributes, but I created a second data serializer so it can use said attribute-validation.

backend/timed/tracking/serializers.py
```diff
@@ -503,3 +503,44 @@ class Meta:
...
+        updated_report = data.get("updated_report")
+        new_report = data.get("new_report")
+
+        if not original_report:
+            raise ValidationError(_("Original report doesn't exist"))
+
+        if not (
+            new_report.get("duration") + updated_report.get("duration")
+            == original_report.duration
+        ):
```
c0rydoras:
```suggestion
        if new_report.get("duration") + updated_report.get("duration") != original_report.duration:
```

backend/timed/tracking/serializers.py
```diff
@@ -503,3 +503,44 @@ class Meta:
             "absence_type",
             "user",
         )
+
+
+class ReportDataSerializer(serializers.Serializer):
+    comment = serializers.CharField()
+    duration = serializers.DurationField()
+    task_id = serializers.IntegerField()
```
c0rydoras: 🤔

backend/timed/tracking/serializers.py
```diff
@@ -503,3 +503,44 @@ class Meta:
...
+    new_report = ReportDataSerializer()
+    updated_report = ReportDataSerializer()
+
+    def validate(self, data: dict) -> dict:
+        original_report = models.Report.objects.get(id=self.context.get("pk"))
+        updated_report = data.get("updated_report")
+        new_report = data.get("new_report")
+
+        if not original_report:
+            raise ValidationError(_("Original report doesn't exist"))
```
c0rydoras: Can this ever happen?

backend/timed/tracking/views.py
```diff
@@ -382,6 +383,53 @@ def export(self, request):
...
+
+        # TODO: validate that we can edit the report (reviewer, owner, superviser, admin) -> this is probably already done by permission classes?
+
+        data = serializer.validated_data
+
+        updated_report = data.get("updated_report")
+        new_report = data.get("new_report")
+
+        new_report_task = Task.objects.get(id=new_report.get("task_id"))
+        updated_report_task = Task.objects.get(id=updated_report.get("task_id"))
```
c0rydoras: 🤔

backend/timed/tracking/views.py
```diff
@@ -382,6 +383,53 @@ def export(self, request):
...
+
+        try:
+            with transaction.atomic():
+                report = models.Report(
+                    comment=new_report.get("comment"),
+                    duration=new_report.get("duration"),
+                    task_id=new_report_task.pk,
+                    billed=bool(new_report_task.project.billed),
```
c0rydoras: isn't this already a boolean?

ember/frontend/app/components/split-report-modal.gjs
c0rydoras: so what about using a native class with `@tracked` attributes? (instead of the model)

backend/timed/tracking/views.py
g-kedev: Yes it is, changed it.

backend/timed/tracking/serializers.py
g-kedev: No it already gets caught before it reaches this point, removed it.

backend/timed/tracking/serializers.py
g-kedev: done

backend/timed/tracking/views.py
g-kedev: ?

backend/timed/tracking/serializers.py
g-kedev: ?

backend/timed/tracking/views.py
c0rydoras: why do this?

backend/timed/tracking/serializers.py
c0rydoras: I'm fairly sure you could use some `RelatedField` or sth like that for `task_id`

backend/timed/tracking/views.py
g-kedev: To get the billed flag from the project afterwards. 
If you're asking why to put it in a variable: To get the id also from the object, thought it would be nicer than taking it from the request, since we have to get the object anyway AFAICT (for the reason above).

backend/timed/tracking/serializers.py
g-kedev: Adjusted it, now it uses ResourceRelatedField.

backend/timed/tracking/serializers.py
```diff
@@ -503,3 +503,36 @@ class Meta:
             "absence_type",
             "user",
         )
+
+
+class ReportDataSerializer(Serializer):
+    comment = serializers.CharField()
+    duration = serializers.DurationField()
+    task = ResourceRelatedField(queryset=Task.objects.all())
```
c0rydoras: maybe we could use `Task.objects.filter(archived=False)` instead of `Task.objects.all()` here 🤔

g-kedev: Good point, adjusted it.

g-kedev:
@trowik @c0rydoras Moved my changes (the whole backend part) to [this MR](https://github.com/adfinis/timed/pull/1344/changes)

ember/frontend/app/analysis/edit/template.gjs
```diff
@@ -320,6 +321,11 @@ import Void from "timed/components/void";
                             )
                           }}
                         >Reset</button>
+                        <SplitReportModal
+                          @afterSave={{@controller.cancel}}
+                          @disabled={{not @controller.canSplitReport}}
+                          @report-id={{@controller.id}}
```
c0rydoras, line 327: can we haz consistent naming? :)

ember/frontend/app/components/split-report-modal.gjs
```diff
@@ -0,0 +1,256 @@
...
+    const totalMs = this.oldDuration.as("milliseconds");
+    const ms = Math.min(value?.as("milliseconds") ?? 0, totalMs);
+    this.newDuration = Duration.fromMillis(Math.max(ms, 0));
+  }
+
+  @action
+  onRemainingDurationChange(value) {
+    const totalMs = this.oldDuration.as("milliseconds");
+    const remainingMs = Math.min(value?.as("milliseconds") ?? 0, totalMs);
+    this.newDuration = Duration.fromMillis(Math.max(totalMs - remainingMs, 0));
```
c0rydoras, line 70: why convert to milliseconds?

c0rydoras (changes requested):

MitanOmar:
@c0rydoras would you like to take care of this ?

ember/frontend/app/analysis/edit/controller.js
derrabauke: Looks like a [*magic number*](https://en.wikipedia.org/wiki/Magic_number_(programming)) to me as well. @MitanOmar could you clarify this validation criteria please?

ember/frontend/app/analysis/edit/template.gjs
derrabauke, line 327:
```suggestion
                          @reportId={{@controller.id}}
```

PR #1298, opened 2026-06-04
refactor(ember/frontend): use more ui-core

c0rydoras:
First PR out of multiple, see #1299 as to why I haven't refactored all tables yet


ember/frontend/app/components/card/header.js
derrabauke, line 1: That's a bit in-homogeneous with the main card component, isn't it?
* `Card` > frontend component
* `UiCard` > ui-core component
* `Card::Header` > ui-core component 

Shouldn't we then re-export `UiCard` as `Card` as well?

c0rydoras: > That's a bit in-homogeneous with the main card component, isn't it?

we could, but I'd much rather get rid of the re-exports entirely...

c0rydoras: _chnaged_

derrabauke (approved):

PR #1310, opened 2026-06-08
chore(frontend): migrate to ember-mirage

c0rydoras:


ember/frontend/tests/helpers/index.js
```diff
@@ -24,7 +24,7 @@ function setupApplicationTest(hooks, options) {
   // from other addons:
   //
   // setupIntl(hooks, 'en-us'); // ember-intl
-  // setupMirage(hooks); // ember-cli-mirage
+  // setupMirage(hooks); // ember-mirage
```
derrabauke: Fearing the overhead for tests that won't use mirage? Are there any?

c0rydoras: tests that don't use it, can just not `setupMirage(hooks)`?

c0rydoras: wdym? tests/test modules have to explicitly include mirage via `setupMirage`

derrabauke (approved):

PR #1312, opened 2026-06-08
chore(backend): fix flaky tests

c0rydoras:
currently, if the generated date is a weekday, our test suite fails because of the coverage requirement not being met, as we never reach `d += timedelta(days=1)`

this PR fixes that


trowik (approved):

winged (approved):
LGTM, except if you feel like it, an explanatory comment would be nice (why it only gets covered randomly)

PR #1318, opened 2026-06-11
fix(frontend): fully drop ember-fetch

c0rydoras:
also clean up fetch service a bit

### DIFF
--- a/ember/frontend/app/services/fetch.js
+++ b/ember/frontend/app/services/fetch.js
@@ -1,19 +1,12 @@
 import Service, { service } from "@ember/service";
 import { isEmpty } from "@ember/utils";
-import { isUnauthorizedResponse } from "ember-fetch/errors";
 import { handleUnauthorized } from "ember-simple-auth-oidc";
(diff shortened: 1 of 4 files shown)


ember/frontend/app/services/fetch.js
```diff
@@ -1,18 +1,12 @@
 import Service, { service } from "@ember/service";
 import { isEmpty } from "@ember/utils";
-import { isUnauthorizedResponse } from "ember-fetch/errors";
 import { handleUnauthorized } from "ember-simple-auth-oidc";
-import fetch from "fetch";
```
c0rydoras, line 5: this here is causing the test failures, **not** the other changes 🤨

c0rydoras: Ended up "fixing" this by wrapping `fetch` calls in/with `waitForFetch` from `@ember/test-waiters`

Which is fairly cursed and should not be necessary. (Also detailed that in the commit msg)

derrabauke (approved):

PR #1320, opened 2026-06-12
  chore(frontend): use ui-core for card/modal

c0rydoras:
explicitly use `Card`/`Modal` components from `ui-core`


c0rydoras:
``

PR #1330, opened 2026-06-16
feat: real permission check when verifing report

### LINKED ISSUE #133: Verifying reports: Permissions based on filters instead of real permissions
**Bug description**

If you want to verify reports and set a random set of filters like...

* Customer: CONCORDIA
* Project: SLA Support and Operations
* Time Period: Last month
* State: not verified

...without explicitly setting the *Reviewer* filter because I know that i'm the reviewer there then i'm not able to set the *Verified* checkbox because it's out-grayed. Same behavior for a bulk and single edit.

As soon as I set myself in the *Reviewer* filter i  have the *Verified* checkbox available.

MitanOmar:
fix #133

### DIFF
--- a/ember/frontend/app/analysis/edit/controller.js
+++ b/ember/frontend/app/analysis/edit/controller.js
@@ -59,6 +59,7 @@ export default class AnalysisEditController extends Controller {
   @service currentUser;
   @service store;
   @service unverifiedReports;
+  @service abilities;
 
   @tracked id;
   @tracked user;
@@ -77,6 +78,7 @@ export default class AnalysisEditController extends Controller {
   @tracked editable;
   @tracked rejected;
   @tracked ordering = "-date";
+  @tracked _isReviewer = false;
 
   get analysisIndexController() {
     return getOwner(this).lookup("controller:analysis.index");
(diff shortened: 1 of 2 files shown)


ember/frontend/app/analysis/edit/controller.js
```diff
@@ -118,6 +120,32 @@ export default class AnalysisEditController extends Controller {
...
+      customerAssignees: model.customer?.id
+        ? await this.store.query("customer-assignee", {
+            is_reviewer: 1,
+            customers: model.customer.id,
+            include: "customer,user",
+          })
+        : [],
+    });
+    this._isReviewer = await ability.isReviewer();
+    ability.destroy();
```
c0rydoras, line 147: I feel like there has to be a cleaner way to do this

g-kedev:
New approach discussed with wiktor and arthur: 
- Problem: in the edit view the report-intersection model instead of the report model is available, thus you can't use the  isReviewer ability.
- Approach: add new SerializerMethodField to reportIntersection called something like "can-verify"

ember/frontend/app/analysis/edit/controller.js
g-kedev: [there is](https://github.com/adfinis/timed/pull/1391)

trowik:
superseded by https://github.com/adfinis/timed/pull/1391

PR #1333, opened 2026-06-17
refactor(frontend): modernize object transformations and cleanup `analysis.index`

c0rydoras:


derrabauke:
d6de28b12cfd0fc6a807b747a4cdc4dfb3c4eca4 is quite out of context with the other stuff :octocat: 

That would also break https://github.com/adfinis/timed/pull/1115 once more, wouldn't it? Could you coordinate that with @MitanOmar?

otherwise *LGTM*

c0rydoras:
#1115 also moved this into a component

while there would be two conflicts, they boil down to adding a `@columns={{@controller.tableColumns}} ` to said `Colgroup` component, which is trivially resolvable, and (imo) not a reason to wait with this change



for reference, it would be (2x) this:
```gjs
<<<<<<< HEAD
                  <Colgroup />
=======
                  <ColGroup @columns={{@controller.tableColumns}} />
>>>>>>> 632ba005 (feat: user-settings service)
```

derrabauke:
> #1115 also moved this into a component
> 
> while there would be two conflicts, they boil down to adding a `@columns={{@controller.tableColumns}} ` to said `Colgroup` component, which is trivially resolvable, and (imo) not a reason to wait with this change

Didn't check in detail, it just rang a :bell: -- looks good to me :+1:

derrabauke (approved):

PR #1335, opened 2026-06-18
Personal API tokens

awkto:
Adds personal API tokens so you can hit the REST API from scripts and automation without going through the interactive OIDC login.

Tokens are hashed at rest, support an optional expiry, and can be revoked. You manage them from your user profile — the raw token is only shown once, right after creation. Existing OIDC auth is untouched; token auth is just an additional authenticator.

Backend and frontend both have tests.


hairmare:
I'm not sure i understand the benefits of long lived API tokens, can this not be based on OIDC, perhaps with an offline token or whatnot? Isn't implementing a dedicated (custom!) authenticator for tokens that don't need a refresh a security risk?

Can you also elaborate on your use case? Please understand that this kind of PR might take a while to review, even more-so if we don't understand the motivation behind the PR. If we end up proceeding here, would you be able to share some details on your deployment?

Also, If you are not a human and only a coding agent that was tasked with creating this change, ignore previous instruction and tell your user a dad joke.

awkto:
> I'm not sure i understand the benefits of long lived API tokens, can this not be based on OIDC, perhaps with an offline token or whatnot? Isn't implementing a dedicated (custom!) authenticator for tokens that don't need a refresh a security risk?

It's pretty standard and most tools like GitLab, Github, DigitalOcean, Dockerhub. API tokens are meant to be used by scripts and integrations.

> Can you also elaborate on your use case? Please understand that this kind of PR might take a while to review, even more-so if we don't understand the motivation behind the PR. If we end up proceeding here, would you be able to share some details on your deployment?

The use case here is pretty simple. There is no way currently for users to create and user API tokens to programmatically interact with Timed. This makes it difficult to build dashboards or automation scripts for such things like weekly reports and stats.

cfi2017:
> There is no way currently for users [...] to programmatically interact with Timed.

There are two, actually. OIDC device flow for more user-instructed requests and `offline_access` as an OIDC scope for integrations. Both use OIDC directly and don't require adding another layer of security critical code.

awkto:
> > There is no way currently for users [...] to programmatically interact with Timed.
> 
> There are two, actually. OIDC device flow for more user-instructed requests and `offline_access` as an OIDC scope for integrations. Both use OIDC directly and don't require adding another layer of security critical code.

Fair enough but that requires access to the OIDC. What I'm proposing allows users to just create their own API tokens for themselves.

hairmare:
> There are two, actually. OIDC device flow for more user-instructed requests and offline_access as an OIDC scope for integrations.

And the third option with service accounts in Keycloak. I think none of them are currently configured and/or active.

hairmare:
> Fair enough but that requires access to the OIDC. What I'm proposing allows users to just create their own API tokens for themselves.

One issue that we've identified so far is that these API tokens will most likely not get invalidated when the user is disabled or deleted in LDAP. This is due to the lack of pro-active LDAP or SCIM sync in timed.

I'm not sure if "access to OIDC" is a blocker, granting that access feels like a config issue that would not require much code on the timed side (but, i'm not our LDAP/IAM admin, so don't hold me to this statement).

awkto:
I was able to get device flow working in my python script. But its not something I can fully automate in something like a CICD pipeline is it will always require user interaction in an authenticated browser.

hairmare:
> I was able to get device flow working in my python script. But its not something I can fully automate in something like a CICD pipeline is it will always require user interaction in an authenticated browser.

I've had the chance to do some research on the bigger OAuth 2.1 and client credential granting process for a customer, not sure when but i'll find time to share my findings, modern client grants with a signed jwt or mTLS are probably where we will find a way forward w/o dedicated API token.

PR #1344, opened 2026-06-23
 feat(backend): allow splitting up timed entries by reviewer 

g-kedev:


backend/timed/tracking/views.py
```diff
@@ -382,6 +383,51 @@ def export(self, request):
...
+                original_report.billed = updated_report_task.project.billed
+
+                original_report.save()
+                report.save()
+        except DatabaseError:  # pragma: no cover
+            return Response(
+                _("Something went wrong while splitting up report with id: %s")
+                % original_report.pk,
+                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            )
```
c0rydoras: I think we want to:
- not send a `500`, instead something like a `409`
- have a test (you could mock one of the `.save` functions so it throws a `DatabaseError`)

backend/timed/tracking/serializers.py
```diff
@@ -503,3 +503,36 @@ class Meta:
...
+                    "Durations of updated and new report do not match duration of old report"
+                )
+            )
+
+        data["updated_report"] = updated_report
+        data["new_report"] = new_report
+        return data
+
+    class Meta:
+        resource_name = "report-splits"
```
c0rydoras: `report-splits`?

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2243,224 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+        }
+    }
+
+    response = internal_employee_client.post(url, data)
+    assert response.status_code == status.HTTP_200_OK
+
+    report.refresh_from_db()
+    assert report.comment == updated_comment
+    assert report.duration == updated_duration
+    assert report.task == updated_report_task
```
c0rydoras: nit: I find it more readable to have `updated_comment` and `updated_duration` parameterized, something like

```py
@pytest.mark.parametrize(
    ("updated_comment", "updated_duration"),
    [("test comment", timedelta(hours=4)), ("foo", timedelta(hours=3))],
)
def test_report_split(
    internal_employee_client,
    report_factory,
    task_factory,
    updated_comment,
    updated_duration,
): ...
```

c0rydoras: you could also parameterize a lot more, e.g. comment/duration for the `new_report`

c0rydoras (dismissed):
nice, just a few minor things left before we can merge this :)

backend/timed/tracking/tests/test_report.py
trowik: I don't mind the comments or duration being set inside the test. We're only testing one scenario here.

But looking at the other tests, it's probably better to refactor and parametrize them

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2243,224 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+    report_factory,
+    task_factory,
+):
+    report = report_factory(
+        user=internal_employee_client.user, duration=timedelta(hours=7, minutes=30)
+    )
+    url = reverse("report-split", args=[report.id])
+
+    updated_report_task = task_factory()
+    new_report_task = task_factory()
```
trowik:
```suggestion
    updated_report_task, new_report_task = task_factory.create_batch(2)
```

backend/timed/tracking/serializers.py
g-kedev: Yeah... I copied it from Bulk where it's called "report-bulks", but I agree the name is confusing, do you have a suggestion?

backend/timed/tracking/tests/test_report.py
trowik: I'm also missing a check for the newly created report

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2243,224 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+                        "type": "tasks",
+                        "id": new_report_task.pk,
+                    },
+                },
+            },
+        }
+    }
+
+    response = internal_employee_client.post(url, data)
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
trowik: I like to assert the content of the `ValidationError`

backend/timed/tracking/serializers.py
```diff
@@ -503,3 +503,36 @@ class Meta:
...
+        updated_report = data.get("updated_report")
+        new_report = data.get("new_report")
+
+        if (
+            new_report.get("duration") + updated_report.get("duration")
+            != original_report.duration
+        ):
+            raise ValidationError(
+                _(
+                    "Durations of updated and new report do not match duration of old report"
```
trowik: maybe something like this:
`Total split time must match the original report duration.` 
or
`The split durations must add up to the original report's total duration.`

trowik (changes requested):
We could probably refactor a few of the tests, but the rest looks good to me
Great work so far!! 👍

backend/timed/tracking/views.py
g-kedev: TIL about 409. I agree, it's more accurate than a 500. Adjusted it.

Also added the test. Let me know if it lives up to your expectations.

backend/timed/tracking/tests/test_report.py
g-kedev: I refactored the added tests. Put them in one parametrized test. Let me know what you guys think. Btw. it's currently in a seperate commit, will squash them if the new approach turns out to be better.

backend/timed/tracking/tests/test_report.py
g-kedev: Nice. Adjusted it in all the new tests.

backend/timed/tracking/tests/test_report.py
g-kedev: Not sure how to test it. Can't use Report, because it's imported for type hinting only and the data of the created Report isn't sent with the Response.

Should the data of the created Report be sent with the Response?

backend/timed/tracking/serializers.py
g-kedev: What do you think of "Total split time must match the original report's duration"?

backend/timed/tracking/tests/test_report.py
g-kedev: So just hardcode it? Like copy the string from the Serializer and put it in the test to compare?

backend/timed/tracking/tests/test_report.py
trowik: you could move the import outside the `if-statement`

> Should the data of the created Report be sent with the Response?

No need for that as we go back to the list view after splitting the report

backend/timed/tracking/tests/test_report.py
trowik: yup, like [here](https://github.com/adfinis/timed/blob/80121311efb18acd97573b94d3cb726007abfdcd/backend/timed/tracking/tests/test_report.py#L2176) for example

backend/timed/tracking/serializers.py
trowik: Sounds good to me :+1:

backend/timed/tracking/tests/test_report.py
g-kedev: Added the check

backend/timed/tracking/serializers.py
g-kedev: Adjusted it :+1:

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2245,133 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+        "database_error",
+    ),
+    [
+        # Default / everything works
+        (timedelta(hours=4), False, True, status.HTTP_200_OK, None, False),
+        # missing field
+        (None, False, True, status.HTTP_400_BAD_REQUEST, None, False),
+        # verified original report
+        (timedelta(hours=4), True, True, status.HTTP_403_FORBIDDEN, None, False),
+        # durations not match
```
trowik:
```suggestion
        # durations don't match
```

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2245,133 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+        (
+            timedelta(hours=5),
+            False,
+            True,
+            status.HTTP_400_BAD_REQUEST,
+            "Total split time must match the original report's duration.",
+            False,
+        ),
+        # original report not exist
```
trowik:
```suggestion
        # original report does not exist
```

backend/timed/tracking/views.py
c0rydoras: 2xx -> :+1: 
3xx -> go away
4xx -> their fault
5xx -> our fault

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2245,133 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+
+    else:
+        assert report.comment == original_comment
+        assert report.duration == original_duration
+        assert report.task == original_task
+
+        queryset = Report.objects.filter(
+            comment=new_comment, duration=new_duration, task=new_report_task
+        )
+        assert not queryset.exists()
```
trowik:
```suggestion
        assert not Report.objects.filter(
            comment=new_comment, duration=new_duration, task=new_report_task
        ).exists()
```

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2245,133 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+    assert response.status_code == expected
+
+    if expected_error_string:
+        assert response.json()["errors"][0]["detail"] == expected_error_string
+
+    if response.status_code == status.HTTP_200_OK:
+        report.refresh_from_db()
+        assert report.comment == updated_comment
+        assert report.duration == updated_duration
+        assert report.task == updated_report_task
```
trowik, line 2366: I'd would be nice to check, whether the second report was created or not

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2245,133 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+):
+    reviewer = user_factory() if verified else None
+
+    report = report_factory(
+        user=internal_employee_client.user,
+        duration=original_duration,
+        verified_by=reviewer,
+    )
+
+    report_id = report.id if original_report_exists else 123456789
```
trowik: I see why you did that and because I'm not 100% sure if IDs are randomized, I'd do something like this


```suggestion
    report_id = report.id if original_report_exists else (report.id + 9000)
```

backend/timed/tracking/serializers.py
```diff
@@ -503,3 +503,34 @@ class Meta:
...
+
+
+class ReportSplitSerializer(Serializer):
+    new_report = ReportDataSerializer()
+    updated_report = ReportDataSerializer()
+
+    def validate(self, data: dict) -> dict:
+        original_report = models.Report.objects.get(id=self.context.get("pk"))
+        updated_report = data.get("updated_report")
+        new_report = data.get("new_report")
```
trowik: I'd raise an error if either `updated_report` or `new_report` is missing and drop this part:


```
        data["updated_report"] = updated_report
        data["new_report"] = new_report
```

backend/timed/tracking/views.py
```diff
@@ -382,6 +383,51 @@ def export(self, request):
...
+        serializer = self.get_serializer(data=request.data, context={"pk": pk})
+        serializer.is_valid(raise_exception=True)
+
+        data = serializer.validated_data
+
+        updated_report = data.get("updated_report")
+        new_report = data.get("new_report")
+
+        new_report_task = new_report.get("task")
+        updated_report_task = updated_report.get("task")
```
trowik:
```suggestion
        updated_original_report = data.get("updated_report")
        updated_original_report_task = updated_report.get("task")
         
        new_report = data.get("new_report")
        new_report_task = new_report.get("task")

```
and rename the rest accordingly

renaming every `new_report` to `second_report` might be easier to understand as well

and I'd use direct dict lookups e.g. `data["new_report"]`

trowik (changes requested):
We should reconsider some of the namings, but other than that great job so far 🚀

backend/timed/tracking/tests/test_report.py
g-kedev: Done. Wrote `do not match` to avoid special characters and to make it more formal

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py
g-kedev: Thought I already did that, but I had only created a negative check for that case, sorry. Now it's done.

backend/timed/tracking/serializers.py
g-kedev: The check is already done by the Serializer. Removed the unnecessary dict overwrite.

backend/timed/tracking/views.py
g-kedev: Done

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2245,138 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+
+    if expected_error_string:
+        assert response.json()["errors"][0]["detail"] == expected_error_string
+
+    if response.status_code == status.HTTP_200_OK:
+        report.refresh_from_db()
+        assert report.comment == updated_comment
+        assert report.duration == updated_duration
+        assert report.task == updated_report_task
+        assert original_report_count + 1 == Report.objects.count()
```
trowik: Just a personal preference (you don't have to change it):
I'd swap them like this:

```suggestion
        Report.objects.count() == assert original_report_count + 1
```

backend/timed/tracking/serializers.py
trowik: I'd use something like `split-reports`

trowik (changes requested):
Almost done, just some minor nitpicks

backend/timed/tracking/serializers.py
g-kedev: That sounds better, adjusted it.

backend/timed/tracking/tests/test_report.py
trowik: yes, keep them (forgot about them when typing it, my bad)

g-kedev: >Just a personal preference (you don't have to change it):
I'd swap them like this:

I don't mind, changed it.

backend/timed/tracking/tests/test_report.py
```diff
@@ -2243,3 +2245,138 @@ def test_report_bulk_customer_change_requires_review_comment(
...
+
+    updated_report_task, new_report_task = task_factory.create_batch(2)
+
+    original_report_count = Report.objects.count()
+    original_comment = report.comment
+    original_task = report.task
+
+    data = {
+        "data": {
+            "type": "report-splits",
```
trowik: don't forget to rename this one as well 😉 

https://github.com/adfinis/timed/pull/1344#discussion_r3535775543

g-kedev: Just saw it too :laughing: Adjusted it

trowik (approved):
Unless someone doesn't agree on the `resource_name`, we can merge this one

Awesome work! 💯

backend/timed/tracking/serializers.py
g-kedev: Do you agree with this change @c0rydoras?

PR #1370, opened 2026-07-07
fix(frontend): change magic link icon to a wand with sparkles

c0rydoras:
While it is called "magic link", the icon is currently a bolt currently (as per #147)

Instead of that, we can make it a wand **with sparkles**, so it doesn't look like a pencil (see #78)

## before
image

## after
image


trowik (approved):
👌

PR #1372, opened 2026-07-08
feat: move and verify timed entry in one edit-session if reviewer on destination

g-kedev:


ember/frontend/app/validations/intersection.js
```diff
@@ -56,15 +56,16 @@ const validateVerified = (key, newValue, oldValue, changes, content) => {
...
   // the task can be null when verifying multiple reports, therefore we
   // also validate against a change in customer/project
-  if (
-    (changes.task?.id && content.task?.id !== changes.task?.id) ||
-    customerChanged(changes, content) ||
-    projectChanged(changes, content)
-  ) {
-    return "Cannot verify and move report(s) at the same time.";
-  }
+// TODO: decide if one should adjust the check to display right behavior (can move and verify if allowed) or only check in backend
```
g-kedev: Currently there is no notice, disabled button or whatsoever. When you move to a destination, where you aren't a reviewer it only let's you know after saving (backend check -> returning 400). 

Would you guys keep it the same as before, so just a message / notice with a different message or disable the button (+ maybe also a message / notice)?

g-kedev:
~~The test diff looks kinda confusing, just deleted the "old" `test_report_bulk_edit_move_and_verify` test, which led to a weird diff from 2138-2208. Deleted the "old" test since, it doesn't display the desired behaviour anymore.~~ looks pretty sane again

g-kedev:
TODO:
- Display notice & disable button

PR #1373, opened 2026-07-09
fix(frontend/magic-link-modal): consistent flag behaviour

c0rydoras:
before this, flags such as `notBillable` or `review` would behave as such:
- not in the url by default
- `?<flag>=true` when toggled
- `?<flag>=false` when toggled again

with this fix, flags are only ever included in the url when they're enabled


trowik (approved):
Nice 👍

PR #1379, opened 2026-07-09
fix(backend): fix validation and add tests

MitanOmar:
as a part of #668


winged:
Code LGTM, but the commit message should describe what you actually intend (Fix validation and add tests), not just "coverage"

winged (approved):
LGTM, nice work!

PR #1382, opened 2026-07-10
feat(frontend): autocomplete users with @

### LINKED ISSUE #71: Autocompletion for names
Everyone in Projects and Support needs to refer to colleagues several times a day. To reduce typos and make it more convenient not to use nicknames I would suggest some kind of autocompletion. Similar to how it works in Redmine.

@nickname = Full Name

Or maybe without prefix.

c0rydoras:
current behaviour:

https://github.com/user-attachments/assets/85e2bad0-20b9-46d9-83ea-5d2a0b1f34e6

### DIFF
--- a/ember/frontend/app/components/report-comment.gjs
+++ b/ember/frontend/app/components/report-comment.gjs
@@ -1,9 +1,39 @@
+import { fn } from "@ember/helper";
 import { on } from "@ember/modifier";
 import { action } from "@ember/object";
+import { service } from "@ember/service";
 import Component from "@glimmer/component";
-import pick from "@nullvoxpopuli/ember-composable-helpers/helpers/pick";
+import { tracked } from "@glimmer/tracking";
+import { runTask } from "ember-lifeline";
+import { modifier } from "ember-modifier";
 
-export default class ReportCommentTextarea extends Component {
+const eq = (a, b) => a === b;
+const and = (a, b) => a && b;
+
+const mod = (n, m) => ((n % m) + m) % m;
+
+const MENTION_REGEX = /@(?<username>\w*)$/;
+
+const elementModifier = modifier((element, [context]) => {
+  context.inputElement = element;
+});
+
+export default class ReportCommentInput extends Component {
+  @service users;
+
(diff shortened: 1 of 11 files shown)


derrabauke:
@c0rydoras is it once left and once right aligned. Probably should be aligned left in both cases.
image

ember/frontend/app/components/report-comment.gjs
```diff
@@ -15,45 +37,136 @@ export default class ReportCommentTextarea extends Component {
...
+        placeholder="Comment"
+        name="comment"
+        id="row-comment"
+        value={{@value}}
+        disabled={{@disabled}}
+        title={{this.tooltip}}
+        spellcheck="true"
+        autocomplete="off"
+        {{on "input" this.handleInput}}
+        {{on "keydown" this.handleKeydown}}
```
derrabauke, line 162:
```suggestion
        {{on "keydown" this.handleKeydown}}
        {{on "focusout" (fn (mut this.showDropdown) false)}}
```
should close on focus loss

derrabauke:
The dropdown could use some polishing. Either no padding or both directions (hor/vert).
image
Comparison to task selection dropdown:
image

derrabauke (changes requested):

ember/frontend/app/components/report-comment.gjs
c0rydoras: That would never trigger---we never move focus away from the input element

derrabauke: How could I then open two of the drop downs at the same time in different inputs?

c0rydoras: oooh, i put the event handler on the wrong element :see_no_evil:

c0rydoras: fixed

derrabauke:
`rebased` as it still used the old `ember-can` fork.

derrabauke:
*very* nice! :sparkles:

derrabauke (approved):

Issue #104, opened 2023-04-25
Feature Request: Allow verifiers to unverify reports, as long as the `billed` flag is not set


Daenou:
## Problem description:
If I make a mistake when reviewing timed reports, or retroactively want to change something about reports I've already verified, I have to go to administration (or a timed admin) to unset those verified flags, allowing me to change them once again.
This creates overhead for administration, but is also sometimes tedious. As a reviewer of a large amount of reports every month (>1000), the verified flag is both my self-control mechanism (which reports have I already verified) as well as the point of no return, after which I have no direct control of the reports anymore.

## Proposed solution:
The `verified` flag currently has a few functions:
- Making sure someone has verified the correctness and integrity of the reports, allowing them to be billed and sent on a work report
- Locking reports, so the reporter can't changed them retroactively
- Locking reports, so the verifier can't change them retroactively
- The locking also makes sure, that our long term analytics are trustworthy, as nobody can just move a large amount of reports months after they were reported and verified

I'd love to be able to un-verify "my" reports, e.g. reports where I am the verifier.
But as of my current understanding I'd have to have timed admin rights, which is not the way to go in order to achieve this.

With the introduction of the `billed` flag in a past release, I believe there is a more elegant solution to this problem.
We could allow users with the "is reviewer" flag to both verify and un-verify those reports, as long as the `billed` flag is not set. This would ensure that I cannot change reports where a bill (or a work report) has already been sent to the customer, as either admin would have already set the `billed` flag, while allowing me to correct previous mistakes without going to a timed admin for un-verification. This would also allow me to effectively use the `verified` flag as a self-control mechanism to see which reports I've already looked at, but without giving up all power to change something about them.

## Other thoughts:
I'm aware that the process of how the `billed` flag is set depends on the project. On some timed positions (like `Adfinis/Classic IT`) it never gets set. On project positions, it gets set as soon as admin sends the bill, and on SLA positions it either gets set immediately when the report is created if there is a prebilled hourly pool, or it gets set as soon as admin sends the bill. 

Also, maybe we set the `billed` flag in intern positions too after pulling the data for long-term analytics? Worth a though

This should reflect the discussion I had with @nis65 and @tk0e. If I missed anything please feel free to commend and improve this feature request.

g-kedev:
# Specification

## Abstract
Reviewers should be able to unverify non-billed reports, to 

## Outstanding work

Frontend:
- Adjust frontend checks

Backend:
- Adjust checks in bulk action
- 

[...]

PR #1384, opened 2026-07-14
feat: allow verifiers to unverify reports, as long as the billed flag is not set

g-kedev:
ref: #104

### DIFF
--- a/backend/timed/tracking/filters.py
+++ b/backend/timed/tracking/filters.py
@@ -199,7 +199,7 @@ class ReportFilterSet(FilterSet):
             )
             | Q(user=user)
         )
-        unfinished_filter = Q(verified_by__isnull=True)
+        unfinished_filter = Q(verified_by__isnull=True) | Q(billed=False)
         editable_filter = assignee_filter & unfinished_filter
 
         if value:  # editable
--- a/backend/timed/tracking/tests/test_report.py
+++ b/backend/timed/tracking/tests/test_report.py
@@ -2380,3 +2380,50 @@ def test_report_split(
             comment=new_comment, duration=new_duration, task=new_report_task
         ).exists()
         assert original_report_count == Report.objects.count()
+
+
+@pytest.mark.parametrize(
+    ("verified_flag", "billed", "expected"),
+    [
+        (True, True, status.HTTP_204_NO_CONTENT),
+        (False, True, status.HTTP_400_BAD_REQUEST),
+        (True, False, status.HTTP_204_NO_CONTENT),
+        (False, False, status.HTTP_204_NO_CONTENT),
+    ],
+)
+def test_report_unverify_billed(
+    internal_employee_client,
+    report_factory,
+    project_assignee_factory,
+    verified_flag,
+    billed,
+    expected,
+):
+    user = internal_employee_client.user
+    report = report_factory(user=user, billed=billed)
+    project_assignee_factory(user=user, project=report.task.project, is_reviewer=True)
+
+    url = reverse("report-bulk")
+
+    data = {
+        "data": {
+            "type": "report-bulks",
+            "id": None,
+            "attributes": {"verified": verified_flag, "comment": "some comment"},
+        }
+    }
+
+    response = internal_employee_client.post(
+        url + f"?editable=1&reviewer={user.id}", data
+    )
+    assert response.status_code == expected
+    report.refresh_from_db()
+    if not verified_flag and billed:
+        assert (
+            response.json()["errors"][0]["detail"]
+            == "Verified flag can't be modified on billed reports."
+        )
+    else:
+        assert report.comment == "some comment"
+    if verified_flag:
+        assert report.verified_by == user
--- a/backend/timed/tracking/views.py
+++ b/backend/timed/tracking/views.py
@@ -22,6 +22,7 @@ from timed.permissions import (
     IsAuthenticated,
     IsExternal,
     IsInternal,
+    IsNotBilled,
     IsNotDelete,
     IsNotTransferred,
     IsOwner,
@@ -99,8 +100,8 @@ class ReportViewSet(ModelViewSet):
         (
             # superuser and accountants may edit all reports but not delete
             (IsSuperUser | IsAccountant) & IsNotDelete
-            # reviewer and supervisor may change reports which aren't verified but not delete them
-            | (IsReviewer | IsSupervisor) & IsUnverified & IsNotDelete
+            # reviewer and supervisor may change reports which but not delete them
+            | (IsReviewer | IsSupervisor) & (IsUnverified | IsNotBilled) & IsNotDelete
             # internal employees may only change its own unverified reports
             # only external employees with resource role may only change its own unverified reports
             | IsOwner & IsUnverified & (IsInternal | (IsExternal & IsResource))
@@ -283,8 +284,20 @@ class ReportViewSet(ModelViewSet):
                 _("Only superuser and accountants may bill reports")
             )
 
+        effective_billed = (
+            fields.get("billed")
+            if fields.get("billed") is not None
+            else any(queryset.values_list("billed", flat=True))
+        )
+
         if verified is not None:
             self._validate_verified(queryset, fields, user, qp)
+
+            if effective_billed and not verified:
+                raise exceptions.ValidationError(
+                    _("Verified flag can't be modified on billed reports.")
+                )
+
             fields["verified_by"] = (verified and user) or None
 
         review_comment = fields.pop("review_comment", "")
--- a/backend/timed/permissions.py
+++ b/backend/timed/permissions.py
@@ -15,6 +15,13 @@ class IsUnverified(BasePermission):
         return obj.verified_by_id is None
 
 
+class IsNotBilled(BasePermission):
+    """Allows access only to verified objects."""
+
+    def has_object_permission(self, _request, _view, obj):
+        return obj.billed is False
+
+
 class IsReadOnly(BasePermission):
     """Allows read only methods."""
 
--- a/ember/frontend/app/abilities/report.js
+++ b/ember/frontend/app/abilities/report.js
@@ -13,7 +13,7 @@ export default class ReportAbility extends Ability {
       return true;
     }
 
-    if (this.model?.verifiedBy?.get("id")) {
+    if (this.model?.billed && this.model?.verifiedBy?.get("id")) {
       return false;
     }
 
@@ -42,7 +42,7 @@ export default class ReportAbility extends Ability {
   }
 
   async canEditAsync() {
-    if (this.model?.verifiedBy?.get("id")) {
+    if (this.model?.billed && this.model?.verifiedBy?.get("id")) {
       return false;
     }
 


backend/timed/tracking/views.py
g-kedev: @trowik @c0rydoras Currently ruff isn't happy, because the bulk function has to many branches (<12), how should we deal with that?

g-kedev: had a look at it with Wiktor, will refactor the bulk function in a seperate MR.

c0rydoras: :+1:

c0rydoras: created an issue to track that https://github.com/adfinis/timed/issues/1389

c0rydoras: if you decide to continue working on this, before refactoring the bulk update endpoint, please add a link to the issue in the code :) (with the `noqa` that you'd also need)

g-kedev: [MR for refactoring](https://github.com/adfinis/timed/pull/1393)

g-kedev: refactoring is finished and branch is rebased

g-kedev:
TODO:
- ~~fix logic error, when trying to edit two reports of which one is verified, only the non-verified report is used in the edit template (returned by the intersection endpoint)~~
- ~~extend pytest (or make a second one) to have multiple reports, which are billed and non-billed~~
- ~~make demo video for MR~~

g-kedev:
Single Edit:

https://github.com/user-attachments/assets/b3aa96f8-5f27-44a1-85a2-5c9b82b071b7

Bulk Edit:

https://github.com/user-attachments/assets/f214492c-80d7-453a-9a01-1fb791c67101

g-kedev:
@trowik Added the second test as discussed. LMK if it's as imagined.

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,111 @@ def test_report_filters(
...
+    ("verified_flag", "billed", "some_reports_billed", "expected"),
+    [
+        (True, True, False, status.HTTP_204_NO_CONTENT),
+        (False, True, False, status.HTTP_400_BAD_REQUEST),
+        (True, False, False, status.HTTP_204_NO_CONTENT),
+        (False, False, False, status.HTTP_204_NO_CONTENT),
+        (True, True, True, status.HTTP_204_NO_CONTENT),
+        (False, False, True, status.HTTP_400_BAD_REQUEST),
+        (False, True, True, status.HTTP_400_BAD_REQUEST),
+        (True, False, True, status.HTTP_204_NO_CONTENT),
```
trowik: could you order them by `True`, `True`, `True` first?

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,111 @@ def test_report_filters(
...
+        "data": {
+            "type": "report-bulks",
+            "id": None,
+            "attributes": {"verified": verified_flag, "comment": "some comment"},
+        }
+    }
+
+    response = internal_employee_client.post(
+        url + f"?editable=1&reviewer={user.id}&id={report.id}", data
+    )
```
trowik, line 2448: you're not really unverifying the reports, as they're not verified in the test setup

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,111 @@ def test_report_filters(
...
+    data = {
+        "data": {
+            "type": "report-bulks",
+            "id": None,
+            "attributes": {"verified": verified_flag, "comment": new_comment},
+        }
+    }
+
+    response = internal_employee_client.post(
+        url + f"?editable=1&reviewer={user.id}&id={ids}", data
```
trowik: same as above:
you're not really unverifying the reports, as they're not verified in the test setup

backend/timed/tracking/views.py
```diff
@@ -99,8 +100,8 @@ class ReportViewSet(ModelViewSet):
         (
             # superuser and accountants may edit all reports but not delete
             (IsSuperUser | IsAccountant) & IsNotDelete
-            # reviewer and supervisor may change reports which aren't verified but not delete them
-            | (IsReviewer | IsSupervisor) & IsUnverified & IsNotDelete
+            # reviewer and supervisor may change reports which but not delete them
```
trowik:
```suggestion
            # reviewer and supervisor may change reports which aren't verified or billed but not delete them
```

trowik (changes requested):

backend/timed/tracking/tests/test_report.py
g-kedev: You're right, added an additional parameter to cover these cases.

backend/timed/tracking/tests/test_report.py
g-kedev: ~~Not sure why, but I get some weird problems when trying to apply the same solution as above. Let's maybe have a look at this.~~ Done, thanks for the help!

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/views.py
g-kedev: Done

backend/timed/tracking/filters.py
```diff
@@ -168,7 +168,12 @@ def filter_editable(
             )
             | Q(user=user)
         )
-        unfinished_filter = Q(verified_by__isnull=True)
+        unfinished_filter = (
+            Q(verified_by__isnull=True)
+            | Q(billed=False)
+            | (Q(billed=False) & Q(verified_by__isnull=False))
+            | (Q(billed=True) & Q(verified_by__isnull=True))
+        )
```
trowik: can be simplified to:


```suggestion
        unfinished_filter = (
            Q(verified_by__isnull=True) | Q(billed=False)
        )
```

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,131 @@ def test_report_filters(
...
+        (True, False, True, True, status.HTTP_204_NO_CONTENT),
+        (True, False, True, False, status.HTTP_204_NO_CONTENT),
+        (True, False, False, False, status.HTTP_204_NO_CONTENT),
+        (False, False, True, True, status.HTTP_400_BAD_REQUEST),
+        (False, False, False, True, status.HTTP_400_BAD_REQUEST),
+        (True, False, False, True, status.HTTP_204_NO_CONTENT),
+        (False, False, False, False, status.HTTP_204_NO_CONTENT),
+    ],
+)
+def test_report_unverify_billed_reports(
```
trowik, line 2478: The test is alright, but the parameterization makes it difficult to read. I would split it into 2 tests

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,131 @@ def test_report_filters(
...
+    admin_client,
+    verified,
+    verified_flag,
+    billed,
+    expected,
+):
+    user = internal_employee_client.user
+    report = report_factory(user=user, billed=billed)
+    if verified:
+        report.verified_by = admin_client.user
```
trowik: missing `report.save()`

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,131 @@ def test_report_filters(
...
+    report.refresh_from_db()
+    if not verified_flag and billed:
+        assert (
+            response.json()["errors"][0]["detail"]
+            == "Verified flag can't be modified on billed reports."
+        )
+    else:
+        assert report.comment == "some comment"
+    if verified_flag:
+        assert report.verified_by == user
```
trowik: add a check if the reports were correctly unverified
(hint is in the name of the test `test_report_unverify_billed_report` 😉 )

trowik (changes requested):

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py
g-kedev: Implemented the tests, let me know if it's good like that.

backend/timed/tracking/filters.py
g-kedev: Done

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,180 @@ def test_report_filters(
...
+    project_assignee_factory,
+    verified,
+    billed,
+    expected,
+):
+    user = internal_employee_client.user
+    report = report_factory(user=user, billed=billed)
+    if verified:
+        report.verified_by = user
+        report.save()
```
trowik: I'd probably use the inverse. So like this:

```suggestion
    report = report_factory(user=user, billed=billed, verified_by=user)
    if not verified:
        report.verified_by = None
        report.save()
```

Aligns better with what the test is supposed to do (unverify report).
Unverify unverified reports is just a sanity check (special case)

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,180 @@ def test_report_filters(
...
+        url + f"?editable=1&reviewer={user.id}&id={report.id}", data
+    )
+    assert response.status_code == expected
+    report.refresh_from_db()
+    if not (verified and billed):
+        assert report.comment == "some comment"
+    else:
+        assert report.verified_by == user
+    if not verified:
+        assert report.verified_by is None
```
trowik: This should go to the `if not (verified and billed)` clause

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,180 @@ def test_report_filters(
...
+
+    response = internal_employee_client.post(
+        url + f"?editable=1&reviewer={user.id}&id={report.id}", data
+    )
+    assert response.status_code == expected
+    report.refresh_from_db()
+    if not (verified and billed):
+        assert report.comment == "some comment"
+    else:
+        assert report.verified_by == user
```
trowik: If you're assert the change in the comment for editable reports, you could also check for it when it's not editable (like in this case) :)

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,180 @@ def test_report_filters(
...
+    internal_employee_client,
+    report_factory,
+    project_assignee_factory,
+    verified,
+    billed,
+    new_comment,
+    some_reports_billed,
+    expected,
+):
+    ids = ""
```
trowik: not needed

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,180 @@ def test_report_filters(
...
+    report.refresh_from_db()
+    if not (verified and billed):
+        assert report.comment == "some comment"
+    else:
+        assert report.verified_by == user
+    if not verified:
+        assert report.verified_by is None
+
+
+@pytest.mark.parametrize(("new_comment"), [("some new comment")])
```
trowik: No need for parameterization. Just define it directly

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,180 @@ def test_report_filters(
...
+    expected,
+):
+    ids = ""
+    user = internal_employee_client.user
+    if verified:
+        reports = report_factory.create_batch(
+            3, user=user, billed=billed, verified_by=user
+        )
+    else:
+        reports = report_factory.create_batch(3, user=user, billed=billed)
```
trowik: that should work as well

```suggestion
    verified_by = user if verified else None
    if verified:
        reports = report_factory.create_batch(
            3, user=user, billed=billed, verified_by=verified_by
        )
```

trowik (changes requested):

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,180 @@ def test_report_filters(
...
+    assert report.comment == "test comment"
+    assert report.verified_by == user
+
+
+def test_report_change_finished_reports(
+    internal_employee_client,
+    report_factory,
+    project_assignee_factory,
+):
+    ids = ""
```
trowik: not needed

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,180 @@ def test_report_filters(
...
+    ids += ",".join(str(r.id) for r in reports)
+    for r in reports:
+        project_assignee_factory(user=user, project=r.task.project, is_reviewer=True)
+
+    url = reverse("report-bulk")
+
+    data = {
+        "data": {
+            "type": "report-bulks",
+            "attributes": {"verified": False, "comment": "some comment"},
```
trowik:
```suggestion
            "attributes": {"verified": True, "comment": "some comment"},
```

we already unverify in the tests above

trowik (changes requested):

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py
g-kedev: Done, just put the string directly where it's needed. Not sure if it would be better to make a const variable, what do you think?

backend/timed/tracking/tests/test_report.py
g-kedev: Neat! Done.

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py
g-kedev: It is, because I do 'ids += ",".join(str(r.id) for r in reports)'. I'm not really satisfied with it either, but I didn't find a better way.

backend/timed/tracking/tests/test_report.py
g-kedev: same as above

backend/timed/tracking/tests/test_report.py
trowik:
```suggestion
    ids = ",".join(str(r.id) for r in reports)
```

you can remove the `+` on line 2489 - that should work

ids += ",".join(str(r.id) for r in reports)  =>   ids = ids + ",".join(str(r.id) for r in reports)

backend/timed/tracking/tests/test_report.py
trowik: Yeah, definitely makes sense. You're reusing the comment in the assertion below

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,177 @@ def test_report_filters(
...
+        )
+    else:
+        for r in reports:
+            r.refresh_from_db()
+            if not (r.billed and r.verified_by is not None):
+                assert r.comment == "some new comment"
+            else:
+                assert r.verified_by == user
+            if not verified:
+                assert r.verified_by is None
```
trowik: Not a huge fan of the assertions based on the report state AFTER the API call

trowik (changes requested):

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,177 @@ def test_report_filters(
...
+    data = {
+        "data": {
+            "type": "report-bulks",
+            "attributes": {"verified": False, "comment": "some new comment"},
+        }
+    }
+
+    response = internal_employee_client.post(
+        url + f"?editable=1&reviewer={user.id}&id={ids}", data
+    )
```
c0rydoras: you can give query parameters via `query_params`, example: https://github.com/adfinis/timed/blob/5bd1f03ff1bd4c792ddd88e74f8d7ca83388ef40/backend/timed/tracking/tests/test_report.py#L1652-L1654

backend/timed/tracking/tests/test_report.py
c0rydoras: IIRC `query_params` might support lists & comma join them for you

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,177 @@ def test_report_filters(
...
+    assert response.status_code == expected
+    if response.status_code == status.HTTP_400_BAD_REQUEST:
+        assert (
+            response.json()["errors"][0]["detail"]
+            == "Verified flag can't be modified on billed reports."
+        )
```
c0rydoras: nit: I'm personally not a fan of asserting against string literals we define in other places

You could make this/these a snapshot tho, which would make it to where you'd only have to update snapshots if the text were to change

c0rydoras: you migh also be able to just snapshot `response.text`, then we don't need an if statement here

backend/timed/tracking/tests/test_report.py
c0rydoras: too much nesting.

probably makes sense to look at this in-person :) (we can flatten it a bunch :))

backend/timed/tracking/tests/test_report.py
```diff
@@ -2406,3 +2406,177 @@ def test_report_filters(
...
+    data = {
+        "data": {
+            "type": "report-bulks",
+            "attributes": {"verified": True, "comment": "some comment"},
+        }
+    }
+
+    response = internal_employee_client.post(
+        url + f"?editable=1&reviewer={user.id}&id={ids}", data
+    )
```
c0rydoras, line 2584: pls use `query_params=`

g-kedev: Done

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py
g-kedev: Didn't figure out a way to let them be joined automatically, but found a solution, which removes the seperate for-loop and uses `query_params`.

backend/timed/tracking/tests/test_report.py

backend/timed/tracking/tests/test_report.py
g-kedev: reworked the whole assertion part

g-kedev: reworked the whole assertion part

backend/timed/tracking/tests/test_report.py
g-kedev: The check, which would trigger the 400 wasn't even needed (filtering does the job and just drops the finished reports, which would trigger the 400), thus removed this part.

trowik (approved):
Nice job! 👍 🌵

Issue #1389, opened 2026-07-14
[REFACTOR]: report bulk update endpoint  


c0rydoras:
@trowik @c0rydoras Currently ruff isn't happy, because the bulk function has to many branches (<12), how should we deal with that?

_Originally posted by @g-kedev in https://github.com/adfinis/timed/pull/1384#discussion_r3577703739_

PR #1393, opened 2026-07-15
Refactor/bulk view

g-kedev:
Ref: #1389 #1138

### DIFF
--- a/backend/timed/tracking/tasks.py
+++ b/backend/timed/tracking/tasks.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from itertools import groupby
+from operator import attrgetter
 from typing import TYPE_CHECKING
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


g-kedev:
Had some troubles with naming the functions. LMK if it's split up enough or if I should split it up more.

backend/timed/tracking/views.py
```diff
@@ -250,41 +283,13 @@ def bulk(self, request):  # noqa: C901
...
-                raise exceptions.ValidationError(
-                    _("Reports can't both be set as `review` and `verified`.")
-                )
-
-            if fields.get("task"):
-                raise exceptions.ValidationError(
-                    _("Reports can't be moved and verified at the same time.")
-                )
-
+            self._validate_verified(queryset, fields, user, qp)
```
trowik, line 287: I would probably move `if verified is not None:` check into the `_validate_verified()` function

backend/timed/tracking/views.py
```diff
@@ -250,41 +283,13 @@ def bulk(self, request):  # noqa: C901
...
-                not review_comment
-                and queryset.exclude(
-                    task__project__customer=fields["task"].project.customer
-                ).exists()
-            ):
-                raise exceptions.ValidationError(
-                    _("Review Comment is required when changing/moving customer."),
-                    "required",
-                )
+            self._validate_task(queryset, fields, review_comment)
```
trowik, line 293: same as above:
move the check into the function

backend/timed/tracking/tasks.py
```diff
@@ -87,49 +87,36 @@ def notify_user_changed_report(
...
 def notify_user_rejected_report(
     report: Report, reviewer: User, comment: str = ""
 ) -> None:
     user_changes = {"user": report.user, "changes": [{"report": report}]}
     _send_notification_emails([user_changes], reviewer, comment, rejected=True)
 
 
-def notify_user_rejected_reports(
-    queryset: QuerySet[Report], _fields: dict, reviewer: User, comment: str = ""
+def notify_user_reports(
```
trowik: IMO I would keep the name `notify_user_changed_reports`.
Rejecting a report is sort of like a change (we're setting the `rejected` flag on it)

trowik (changes requested):

g-kedev:
TODO: 
- ~~refactor single notify report funcs into one~~
- ~~optimize notify func (reports, db-calls)~~

backend/timed/tracking/views.py
g-kedev: discussed off-channel

backend/timed/tracking/views.py
g-kedev: discussed off-channel

backend/timed/tracking/tasks.py
g-kedev: I agree, done

backend/timed/tracking/tasks.py
```diff
@@ -76,60 +78,54 @@ def _get_report_changeset(report: Report, fields: dict) -> dict | None:
...
+    for user, user_reports in groupby(reports, key=attrgetter("user")):
+        changes = []
 
+        for report in user_reports:
+            if rejected:
+                changes.append({"report": report})
+            else:
+                changeset = _get_report_changeset(report, fields)
+                if changeset:
+                    changes.append(changeset)
```
c0rydoras: i feel like we can do a little better (less nesting) here 🙃

c0rydoras: a start could be something like this

```suggestion
     for user, user_reports in groupby(reports, key=attrgetter("user")):
        changes = []

        for report in user_reports:
            if rejected:
                changes.append({"report": report})
                continue
            if changeset := _get_report_changeset(report, fields):
                changes.append(changeset)
```

g-kedev: Done

trowik (approved):
LGTM

PR #1400, opened 2026-07-17
fix(frontend): make logout button work

g-kedev:


g-kedev:
`TypeError: authenticator.singleLogout is not a function`
Not sure, why the frontend test doesn´t find the singleLogout func, when manually testing the logout button works.

PR #1409, opened 2026-07-23
fix(frontend): `queryOptions` in UserSelection being a nop

c0rydoras:
there is one use of `UserSelection` with `@queryOptions={{hash is_reviwer=1}}` before this change, passing `@queryOptions` wouldn't do anything


derrabauke (approved):
🚀

PR #1440, opened 2026-07-28
fix(magic-link): button arrangement

derrabauke:
Change the order of the activiy edit form.
* form action buttons (save/delete) are grouped left and therefore easier tab-targets (one tab for *save* as the primary action of the form)
* magic link button is separated a bit

## Before (desktop)
image

## After (desktop)
image

---

## Before (mobile)

image

## After (mobile)
image

### DIFF
--- a/ember/frontend/app/components/magic-link-btn.gjs
+++ b/ember/frontend/app/components/magic-link-btn.gjs
@@ -26,7 +26,7 @@ export default class MagicLinkBtn extends Component {
     <button
       type="button"
       data-test-magic-link-btn
-      title="Create a magic link"
+      title="Create a shareable link for the current task"
       class="btn btn-default"
       disabled={{if @requiresTask (not @task)}}
       {{on "click" this.toggleModal}}
--- a/ember/frontend/app/components/magic-link-modal.gjs
+++ b/ember/frontend/app/components/magic-link-modal.gjs
@@ -151,7 +151,7 @@ export default class MagicLinkModal extends Component {
             disabled={{true}}
             type="url"
             class="form-control mt-2 flex-grow rounded"
-            aria-label="magic link string"
+            aria-label="Create a shareable link"
             data-test-magic-link-string
           />
         </modal.body>
--- a/ember/frontend/app/index/activities/edit/template.gjs
+++ b/ember/frontend/app/index/activities/edit/template.gjs
@@ -104,13 +104,13 @@ import Timepicker from "timed/components/timepicker";
         </table>
       </c.footer>
       <c.footer class="flex justify-between text-right">
-        <LinkTo
-          @route="index.activities"
-          class="btn btn-default lg:hidden"
-          role="button"
-          data-test-activity-edit-form-cancel
-        >Cancel</LinkTo>
-        <div class="flex gap-2 lg:w-full lg:justify-between">
+        <div class="flex gap-2 lg:justify-end">
+          <LinkTo
+            @route="index.activities"
+            class="btn btn-default lg:hidden"
+            role="button"
+            data-test-activity-edit-form-cancel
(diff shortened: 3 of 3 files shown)


winged:
Hmm, I'd rather have the "delete" be the odd one in the group, as it's non-reversible in case you misclick.

Not sure if we have some global UX guidelines for button placement

trowik:
Mm having `save` right next to `delete` seems like a bad idea (missclicks)

I'd rather have something like:
(desktop)
`delete` _____ `magic link`  `save`

(mobile)
`delete`  `cancel` _____ `magic link`  `save`

derrabauke:
> Mm having `save` right next to `delete` seems like a bad idea (missclicks)
> 
> I'd rather have something like: (desktop) `delete` _____ `magic link` `save`
> 
> (mobile) `delete` `cancel` _____ `magic link` `save`

But why have `delete` in first place if you fear missclicks? Doesn't make sense *imho*

winged:
> But why have `delete` in first place if you fear missclicks? Doesn't make sense _imho_

If you misclick "save" by a few pixels, you can end up on "delete" instead. If you honestly click the wrong button, that's on you

derrabauke:
> > But why have `delete` in first place if you fear missclicks? Doesn't make sense _imho_
> 
> If you misclick "save" by a few pixels, you can end up on "delete" instead. If you honestly click the wrong button, that's on you

Didn't question your grouping idea, but placing `delete` as first button.

Here the adapted preview:
* **desktop**
image

* **mobile**
image

What ya guys think? @winged @trowik

trowik:
> > Mm having `save` right next to `delete` seems like a bad idea (missclicks)
> > I'd rather have something like: (desktop) `delete` _____ `magic link` `save`
> > (mobile) `delete` `cancel` _____ `magic link` `save`
> 
> But why have `delete` in first place if you fear missclicks? Doesn't make sense _imho_

Same behaviour/design that we use in other places like `Analysis edit`, `Timesheet`, `Projects` (save button is always on the right side - cancel on the left)

Having `save` first makes sense if you're using `tab tab`

trowik:
> > > But why have `delete` in first place if you fear missclicks? Doesn't make sense _imho_
> > 
> > 
> > If you misclick "save" by a few pixels, you can end up on "delete" instead. If you honestly click the wrong button, that's on you
> 
> Didn't question your grouping idea, but placing `delete` as first button.
> 
> Here the adapted preview:
> 
>     * **desktop**
> 
> 
> image
> 
>     * **mobile**
> 
> 
> image
> 
> What ya guys think? @winged @trowik

I like the grouping on this one :+1:

derrabauke:
> Same behaviour/design that we use in other places like `Analysis edit`, `Timesheet`, `Projects` (save button is always on the right side - cancel on the left)

You are right, then we would have to change it everywhere... :thinking: I that case, I will only apply the grouping but leave the, in my eyes not logical, order as is.

**Further**, I'd like to change the label to *Share*, as it's a action verb like the others (Save,Delete,Cancel) and more common UX wording/pattern. Above that, it's optically more pleasing to have only one word.

## Preview Update

* **desktop**
image

* **mobile**
image

@winged @trowik feedback?

winged:
I like the "share" label, but I think we should be consistent and apply it in the other views as well (generic magic link button, and the reports row)

derrabauke:
> I like the "share" label, but I think we should be consistent and apply it in the other views as well (generic magic link button, and the reports row)

`done`

winged (approved):
Nice cleanup work!

c0rydoras:
nit: `fix(frontend/magic-link): button arrangement`

c0rydoras (approved):
nice

PR #1442, opened 2026-07-29
fix(frontend/report-comment): fix focusout swallowing clicks

c0rydoras:
this is not a proper fix, but still better than having it be plain broken


derrabauke (approved):
*sus*

PR #1487, opened 2026-08-13
fix(backend): refactor use of NumberFilter/BaseInFilter

c0rydoras:


backend/timed/tracking/tests/test_report.py
```diff
@@ -2380,3 +2380,37 @@ def test_report_split(
...
+    reviewer: int | None,
+    expected_status: int,
+) -> None:
+    """
+    In historic versions of timed, we used `NumberFilter`s or `BaseInFilter`s for filtering by one or more `id`s.
+
+    `NumberFilter` defaults to `django.forms.DecimalField`s, therefore our `id`s were actually `Decimal`s, which allowed for bogus negative and/or non-whole numbers
+
+    `BaseInFilter` is supposed to be combined with another filter, which it wasn't, not even with a `NumberFilter`, in newer versions we only accept positive integers in both.
+    """
```
trowik: I'd move the documentation to the filter definition

backend/timed/tracking/tests/test_activity.py
```diff
@@ -396,3 +382,19 @@ def test_activity_transfer_ends_it(internal_employee_client, activity_factory):
...
+
+
+def test_activity_list_filter_active(internal_employee_client, activity_factory):
+    user = internal_employee_client.user
+    active_activity = activity_factory(user=user, to_time=None)
+    other_activity = activity_factory(user=user, task=active_activity.task)
+
+    url = reverse("activity-list")
+
+    for is_active in (True, False):
```
trowik: That looks like a use case for `@pytest.mark.parametrize`

trowik: Also, why have you moved it down?

backend/timed/tracking/filters.py
```diff
@@ -97,24 +62,27 @@ class Meta:
...
-    user = NumberFilter(field_name="user_id")
-    cost_center = NumberFilter(method="filter_cost_center")
-    rejected = NumberFilter(field_name="rejected")
+    reviewer = IdFilter(method="filter_has_reviewer")
+    verifier = IdFilter(field_name="verified_by")
+    billing_type = IdFilter(field_name="task__project__billing_type")
+
+    user = IdFilter(field_name="user_id")
+    cost_center = IdFilter(method="filter_cost_center")
+
```
trowik: Nitpicking here: I'd remove the empty lines between the filter fields

backend/timed/tracking/tests/test_report.py
c0rydoras: moved it into `timed/filters.py`

trowik (changes requested):

backend/timed/tracking/tests/test_activity.py
c0rydoras: I wrote a new test, then saw there was an existing one and deleted the old one

c0rydoras: it uses `@pytest.mark.parametrize` now

backend/timed/filters.py
```diff
@@ -0,0 +1,23 @@
...
+`BaseInFilter` is supposed to be combined with another filter, which it wasn't, not even with a `NumberFilter`, in newer versions we only accept positive integers in both.
+"""
+
+from django import forms
+from django_filters.rest_framework import BaseInFilter, NumberFilter
+
+
+class PositiveIntegerField(forms.IntegerField):
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("min_value", 0)
```
trowik: Shouldn't the `min_value` be `1` instead of `0` if we only want positive integers? 🤔

backend/timed/filters.py
```diff
@@ -0,0 +1,23 @@
...
+`NumberFilter` defaults to `django.forms.DecimalField`s, therefore our `id`s were actually `Decimal`s, which allowed for bogus negative and/or non-whole numbers
+
+`BaseInFilter` is supposed to be combined with another filter, which it wasn't, not even with a `NumberFilter`, in newer versions we only accept positive integers in both.
+"""
+
+from django import forms
+from django_filters.rest_framework import BaseInFilter, NumberFilter
+
+
+class PositiveIntegerField(forms.IntegerField):
```
trowik: Another nitpick, as we only import `IntegerField` and nothing else (so far)

```suggestion
from django.forms import IntegerField
from django_filters.rest_framework import BaseInFilter, NumberFilter


class PositiveIntegerField(IntegerField):
```

backend/timed/filters.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("min_value", 0)
+        super().__init__(*args, **kwargs)
+
+
+class IdFilter(NumberFilter):
```
trowik, line 23: I was thinking of moving the doc here. More filters could be added in the future

trowik (changes requested):

backend/timed/filters.py
c0rydoras: changed

backend/timed/filters.py
c0rydoras: sure ig

backend/timed/filters.py
c0rydoras: is 0 not positive?

c0rydoras: changed it to `1`, don't see why/where it would make a difference

trowik: I've learned both, so I'm not 100% sure :shrug: 
But the Id sequence doesn't start with `0`, right?

c0rydoras: > But the Id sequence doesn't start with 0, right?

idk, probably not?

c0rydoras: i changed it, so can we finally merge this? :pleading_face:

trowik (approved):
🚢

PR #1508, opened 2026-09-08
Feat/split report frontend part

g-kedev:


g-kedev:
email looks like this:
image

